PR #1, opened 2016-01-12
Make all import paths relative to file location

josephco:


misscs:
Looks good to me. I wish we didn't have use `..` relative paths as it makes it harder to parse the file. Maybe we'll find a better way or it'll improve as we rebundle these files into our module system. For now, this works well!

josephco:
@misscs Yeah, it feels less clean, but at least it should work for now. We'll get there.

PR #5, opened 2016-01-20
Update and simplify show/hide display helpers

josephco:
Remove old helpers
Add new hide on small/large screen helpers

### DIFF
--- a/utilities/_display_helpers.scss
+++ b/utilities/_display_helpers.scss
@@ -33,29 +33,16 @@
   @extend %is-hidden;
 }
 
+.hide-on-small-screens {
 
-.is-hidden-on-mobile {
-  display: none !important;
-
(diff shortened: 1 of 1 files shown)


utilities/_display_helpers.scss
```diff
@@ -33,29 +33,16 @@
...
-    display: block !important;
-  }
-}
-
-
-.show-on-mobile {
-  display: inline-block !important;
-
-  @include respond-at-and-above($break-small) {
+  @include respond-below($break-small) {
```
victorwhy, line 34: did we make sure at exactly 600px that it's covered?

josephco, line 34: think we should be fine

josephco:
Verbal DTM from @misscs

PR #6, opened 2016-01-25
Refactor buttons to use data attribute model

josephco:
Remove links and move to separate file


josephco:
Reopening in a new PR

PR #16, opened 2016-02-15
Update trial warranty blue

victorwhy:
I believe we had changed this before, but `trial-warranty` should be #001856

image


misscs:
Looks good to me. I'll merge this in and make the update.

PR #17, opened 2016-02-15
Move modules into nightshade styles

misscs:
This PR must get merged in before https://github.com/CasperSleep/nightshade/pull/12

This PR introduces a dramatic restructure to nigthshade-styles directory
- Deletes `base/` files
  - Files moved into their respective modules in `modules/`
  - Files that don't have a clear association to a module are in `utilities/`
- Reorganizes `modules/` to be a folder with files, which includes macros and JS
- Removed references to files/modules that aren't actively used by Nightshade
## Considerations

There's great inconsistency in organization within modules. I'm hoping we can collectively establish a pattern after we work with the files. The discussion is happening https://github.com/CasperSleep/nightshade-styles/issues/19.


misscs:
@josephco Thanks to the wonders of `git rerere`, updating all our branches was a breeze. You should check to make sure everything from your end behaves as intended. From my review of the commit history it looks :ok_hand:

PR #20, opened 2016-02-18
Sheets product page modules

misscs:
This PR implement modules to render sheets product page. Location of files and filenames are not final. Input is needed from the team to finalize conventions. 

In this PR I did start to use the convention `module-name.macros.html` to denote a file of macros. I like the explicitness over the `module-name/macros.html` convention we were experimenting with. 

This PR must get merged before the update to Nightshade _put PR link here_


misscs:
Whoops. Issued against the wrong branch. Closing

PR #21, opened 2016-02-18
Sheets product page modules

misscs:
This PR implement modules to render sheets product page. Location of files and filenames are not final. Input is needed from the team to finalize conventions. 

**This PR must get merged before the update to Nightshade https://github.com/CasperSleep/nightshade/pull/17**

**Highlights**
- Experimenting with JSDoc style comments across our files. What do folks think about it? I like the idea of having one documentation style. Sassdoc slightly breaks it because it requires that doc lines start with `///`
  - CSS 48cea1e0eafdb
  - Macros 42e1813, 4806fdc9464ec
- Added Imgix. [Added a todo](https://trello.com/c/dimxTePt) to make it part of Ando. 
- Started to use the convention `module-name.macros.html` to denote a modules macros. I like the explicitness over the `module-name/macros.html` convention we were experimenting with. I created [an issue](https://github.com/CasperSleep/nightshade-core/issues/22) to go through this repo and update all macro files to this convention


modules/products/sheets/_a_la_carte.scss
```diff
@@ -98,21 +106,98 @@
...
+    width: 100%
+  }
+
+  .duvet-label-title {
+    font-weight: 600;
+    position: relative;
+    cursor: pointer;
+
+    // custom radio
+    &:before {
```
josephco: this should use the new before() mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -446,3 +498,261 @@
...
+  margin: 0 auto;
+
+ &.is-active{
+  width: 100%;
+ }
+
+
+  &--set {
+
+    &:before {
```
josephco: should use new mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -446,3 +498,261 @@
...
+      box-shadow: (-2px) 2px 0 color(background);
+
+      @include respond-at-and-above($break-large) {
+        box-shadow: (-2px) 2px 0 color(light-background);
+      }
+    }
+
+    &.is-active {
+
+      &:before {
```
josephco: should use new mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -274,7 +325,8 @@
   &--collapsed {
     cursor: pointer;
 
-    @include after($content: "+") {
+    @include after("+") {
```
josephco: This should use the new method (which was removed)

modules/products/sheets/_a_la_carte.scss
```diff
@@ -289,7 +341,7 @@
     }
 
     &.slug-casper-duvet-cover {
-      &:after {
```
josephco: this should revert to what was there without mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -306,7 +358,7 @@
         cursor: not-allowed;
       }
 
-      &:after {
+      @include after("+") {
```
josephco: revert to what was there without mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -446,3 +498,261 @@
...
+
+    &.is-active {
+
+      &:before {
+        width: 1px;
+        left: -1px;
+        border-top-left-radius: 0;
+      }
+
+      &:after {
```
josephco: Use new mixin

modules/products/sheets/_a_la_carte.scss
```diff
@@ -446,3 +498,261 @@
...
+          box-shadow: (-2px) 2px 0 color(light-background);
+        }
+      }
+    }
+  }
+
+  &--custom {
+    right: 3px;
+
+    &:after {
```
josephco: Use new mixin

modules/products/sheets/_a_la_carte.scss
josephco: Use new mixin

josephco:
@misscs I like the JSdoc style in the JS/SCSS, but looks a little bulky in the html/nunjucks templates. Not against it though.

modules/products/sheets/_a_la_carte.scss
```diff
@@ -275,6 +326,7 @@
     cursor: pointer;
 
     @include after($content: "+") {
+      position: absolute;
```
josephco: This line will be included within the after() mixin so we can remove if we want

josephco:
@misscs DTM once we are all cool with it! Looks good to me.

misscs:
@josephco Regarding doc style, any thoughts on how to document the macros? Especially once we start having lots of params.

modules/products/sheets/_a_la_carte.scss
misscs: Good catch. Will clean that up

josephco:
@misscs I guess the JSDoc style is good for params, maybe it's just the stars before each line that looks visually strange to me when in nunjucks/jinja style comment blocks. I wouldn't mind comment lines just starting without anything in front.

misscs:
@josephco Valid. I'm open to trying it out. I like not having to remember that one file is different, but I also am bothered by the mismatch between jinja's `#` and jsdoc `*`.  

I'll try some comments without the `*` in subsequent PRs for static pages.

josephco:
@misscs cool, I think it might be easier to scan, but we can see what we like

misscs:
@josephco Let's get some more input on it.

PR #29, opened 2016-03-01
Modify fancy arrow parameter for link mixin

edwardsfriedman:
For “back” arrow instead of “forward”

### DIFF
--- a/base/typography/_links_config.scss
+++ b/base/typography/_links_config.scss
@@ -27,19 +27,27 @@ $active: '&:active';
     border-bottom: 1px solid color($hover);
   }
 
-  @if $fancy != "" {
-    &:#{$fancy} {
+  &:before,
+  &:after {
+    color: inherit;
+    font-style: normal;
+    line-height: 1;
+  }
 
-      @if $fancy == "before" {
-        content: "‹";
-        margin-right: 2px;
-      } @else {
-        content: "›";
-        margin-left: 2px;
-      }
-      color: inherit;
-      font-style: normal;
-      line-height: 1;
+  @if $fancy == "before" {
+    &:before {
+      content: "‹";
+      margin-right: 2px;
+    }
+  } @else if $fancy == "after" {
+    &:after {
+      content: "›";
+      margin-left: 2px;
+    }
+  } @else {
+    &:before,
+    &:after {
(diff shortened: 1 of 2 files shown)


josephco:
@edwardsfriedman these changes look good to me. Maybe @misscs will want to verify before merging into core.

edwardsfriedman:
@josephco thanks! @misscs whaddaya think?

ckeisato:
@edwardsfriedman Just checking but you did test this with the links that currently use the `fancy` attribute right?

base/typography/_links_config.scss
```diff
@@ -35,5 +36,13 @@ $active: '&:active';
...
       margin-left: 2px;
     }
+  } @else if $fancy-before {
+    &:before {
+      content: "‹";
+      color: inherit;
+      font-style: normal;
+      line-height: 1;
+      margin-right: 2px;
+    }
```
misscs: There's a lot of duplication here. What if we combine the shared properties into one ruleset?

misscs:
Once this is merged in, you'll need to follow [these steps for tagging](https://github.com/CasperSleep/nightshade-core/blob/master/contributing.md#tagging-your-release). Youre release would be a minor update. 

One casper end, you will need to update the version in shrinkwrap. Running `npm install caspersleep/nightshade-core -S` should take care of updating everything. Holler if you have any issues.

base/typography/_links_config.scss
misscs: Another option could be to have one param `$fancy` and have it accept either `before` or `after`. And then run the checks on whether before or after specified. You could interpolate the value for the psedo-selector

``` sass
   &:#{$fancy} {

     @if $fancy = before {
     content: "‹";
      margin-right: 2px;
    } @else { } 
      color: inherit;
      font-style: normal;
      line-height: 1;
    }


```

edwardsfriedman:
@ckeisato yep!

edwardsfriedman:
@misscs Made the changes! so it only accepts three params and expects "before" or "after" as the third. If the parameter is null or a string other than "before" or "after", no arrow will appear.

misscs:
Noice! This iteration is ++

PR #32, opened 2016-03-02
Bump to 1.6.0

robwierzbowski:
Really this is a breaking change (as seen on Casper.com), but keeping a minor
version as requested in PR 31.


victorwhy:
DTM

PR #33, opened 2016-03-03
Remove modules unused by storefront (Casper repo)

josephco:
This will result in a minor version increase to `1.8.0`


misscs:
@josephco I updated the version number as part of the PR. I did it as an experiment, but I think it could work well that after DTM is given we bump the version. Something to try out :)

PR #35, opened 2016-03-04
Standardize naming and structure

josephco:
This PR accompanies https://github.com/CasperSleep/nightshade/pull/29. This PR should be merged first.
- rename modules to src
- move utilities into src
- all macros follow `modulename.macros.html`
- adding an `lib` folder in the module for files that are not part of the bundle. not all modules will have one.
- all files will have a `_index.scss` that imports sass files
- any module can have `_helpers.scss` or `_config.scss`
- helpers: mixins, functions, and classes.
- config: maps and variables. no output.


misscs:
Start of documentation https://github.com/CasperSleep/frontier-codebook/blob/master/nightshade-core.md

PR #45, opened 2016-03-09
Update color helpers and map

josephco:
This is tied to https://github.com/CasperSleep/nightshade/pull/40, which should be merged after this PR


src/notifications/_base.scss
```diff
@@ -23,7 +23,7 @@
 
 @import "flashes_config";
 
-$flash-blue: #e5ecf4; // Replace with value or color mix from map?
+$flash-blue: #e5ecf4; // Replace with value from map?
```
misscs, line 26: Local vars live in the modules config file

PR #47, opened 2016-03-16
Create Press Quote module and macro

josephco:
This PR is tied to https://github.com/CasperSleep/nightshade/pull/47, which should be merged after this

### DIFF
--- a/src/press_quote/quote.macros.html
+++ b/src/press_quote/quote.macros.html
@@ -1,0 +1,15 @@
+{#
+  * @overview Macros for press quotes.
+#}
+
+{% from '../icons/icons.macros.html' import svg %}
+
+{#
+  * Renders a single press quote with source logo beneath.
+#}
+{% macro press_quote(content, logo, class=null, id=null) %}
+<div class="press-quote{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
+  <p class="press-quote-copy body--quote">“{{ content }}”</p>
+  {{ svg(logo) }}
+</div>
+{% endmacro %}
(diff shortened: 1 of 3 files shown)


src/press_quote/quote.macros.html
```diff
@@ -0,0 +1,15 @@
...
+  * @overview Macros for press quotes.
+#}
+
+{% from '../icons/icons.macros.html' import svg %}
+
+{#
+  * Renders a single press quote with source logo beneath.
+#}
+{% macro press_quote(content, logo, class=null, id=null) %}
+<div class="press-quote {% if class %}{{ class }}{% endif %}" {% if id %}id="{{ id }}"{% endif %}>
```
misscs: Since we'll want to add `class` and `id` to all of our macros, I would recommend creating a `utilities.macros.html` file. We can make a macro that adds `{% if class %}{{ class }}{% endif %}" {% if id %}id="{{ id }}"{% endif %}`

src/press_quote/quote.macros.html
```diff
@@ -0,0 +1,15 @@
+{#
+  * @overview Macros for press quotes.
+#}
+
+{% from '../icons/icons.macros.html' import svg %}
+
+{#
+  * Renders a single press quote with source logo beneath.
+#}
+{% macro press_quote(content, logo, class=null, id=null) %}
```
misscs, line 10: Can you make a utility helper macro for class and id?

josephco, line 10: @misscs Yes making that this morning on a separate PR

misscs, line 10: @josephco I'm adding to cleanup but when we document macros we should be documenting the params in the file, too. Let's treat it the same as JSDoc.

PR #49, opened 2016-03-16
Add Hero macro and update styles

josephco:
This PR is tied to https://github.com/CasperSleep/nightshade/pull/49, which should be merged after this

### DIFF
--- a/src/hero/_base.scss
+++ b/src/hero/_base.scss
@@ -1,18 +1,26 @@
-.banner {
-  padding: 50px 23px;
+/**
+ * @file Hero
+ * @module Hero
+ * @overview Hero and top of page featured hero modules
+ */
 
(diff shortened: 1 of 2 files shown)


src/hero/_base.scss
```diff
@@ -1,22 +1,30 @@
...
+.hero-heading {
   font-size: 30px;
   margin-bottom: 10px;
 
   @include respond-at-and-above($break-medium) {
     font-size: 36px;
   }
-}
\ No newline at end of file
+}
```
misscs, line 30: @josephco For future, let's always resort to our ms values whenever we can especially during prototyping.

josephco, line 30: Good call, I somehow missed this

PR #52, opened 2016-03-18
Create Callout module

josephco:
Includes basic structure for Layout module

### DIFF
--- a/src/callout/callout.macros.html
+++ b/src/callout/callout.macros.html
@@ -1,0 +1,13 @@
+{#
+  * @overview Macros for the Callout module.
+#}
+
+{#
+  * Renders a callout module with heading and layout content beneath
+#}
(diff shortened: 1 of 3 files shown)


src/callout/callout.macros.html
```diff
@@ -0,0 +1,13 @@
...
+
+{#
+  * Renders a callout module with heading and layout content beneath
+#}
+{% macro callout(heading, class=null, id=null) %}
+<div class="callout{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
+  <h2 class="callout-heading -dark">{{ heading }}</h2>
+  {{ caller() }}
+</div>
+{% endmacro %}
```
misscs, line 13: We talked on Friday about wrapping our modules in an `inner` or `wrapper` div. This needs an `inner` wrapper or an outer one.

PR #53, opened 2016-03-21
Module updates and tweaks

josephco:
This should be rebased and merged after #55 and #56

### DIFF
--- a/src/content_panel/panel.macros.html
+++ b/src/content_panel/panel.macros.html
@@ -8,7 +8,7 @@
 {% macro content_panel(heading, image_path, image_description=null, image_position='left', class=null, id=null) %}
 <div class="content-panel{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
   <figure class="content-panel-image content-panel-image--{{ image_position }}">
-    <img src="https://casper.imgix.net/{{ image_path }}" alt="{{ image_description }}">
+    <img class="imgix-fluid" src="https://casper.imgix.net/{{ image_path }}" alt="{{ image_description }}">
   </figure>
   <div class="content-panel-content">
     <div class="content-panel-content-body">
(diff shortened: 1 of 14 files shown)


src/content_panel/panel.macros.html
```diff
@@ -8,7 +8,7 @@
 {% macro content_panel(heading, image_path, image_description=null, image_position='left', class=null, id=null) %}
 <div class="content-panel{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
   <figure class="content-panel-image content-panel-image--{{ image_position }}">
-    <img src="https://casper.imgix.net/{{ image_path }}" alt="{{ image_description }}">
+    <img class="imgix-fluid" src="https://casper.imgix.net/{{ image_path }}" alt="{{ image_description }}">
```
misscs, line 11: You may need to pass in a specific class to target these via imgix.

src/hero/_base.scss
```diff
@@ -7,6 +7,7 @@
 .hero {
   display: flex;
   flex-direction: column;
+  align-items: center;
   justify-content: center;
   min-height: 300px;
```
misscs: It's going to be better for us if we avoid pixel values whenever possible. When we want to set height, let's think about it in terms of viewport units: what is the proportion/ratio of the screen we want the element to take up?

misscs:
Not sure if this is ready for review or if you want intermediate feedback. Went ahead and offered my two cents. I'm DTM pending height fix.

src/hero/_base.scss
josephco: I couldn't find a nice viewport unit, so I'm just going to use a modular scale value for now until I can work with design a bit more on that.

josephco:
@misscs I've replaced all the min-height pixel values with modular scale values that are roughly equivalent. Will merge this once we are cool with footer and backgrounds.

PR #54, opened 2016-03-21
Add Blocks module

josephco:
This is branched off `add-callout-module`, so should be rebased and merged after #52

### DIFF
--- a/src/blocks/blocks.macros.html
+++ b/src/blocks/blocks.macros.html
@@ -1,0 +1,26 @@
+{#
+  * @overview Macros for the Blocks module.
+#}
+
+{% from '../icons/icons.macros.html' import svg %}
+
+{#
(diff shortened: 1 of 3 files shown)


src/blocks/blocks.macros.html
```diff
@@ -0,0 +1,26 @@
...
+  * Renders an individual block with an icon and text beneath
+#}
+{% macro block(content, icon) %}
+<div class="block">
+  <figure class="block-image">
+    {{ svg(icon) }}
+  </figure>
+  <h4 class="block-description">{{ content }}</h4>
+</div>
+{% endmacro %}
```
misscs: I feel like block is too generic a module name here. I also feel that the pattern inside block, commonly called media pattern, is one we'll want to address. For now, let's call this `block_item`

PR #55, opened 2016-03-22
Add footer module

josephco:
This is tied to https://github.com/CasperSleep/nightshade/pull/58, which should be merged after this
This should be merged before #53

### DIFF
--- a/src/footer/_base.scss
+++ b/src/footer/_base.scss
@@ -1,0 +1,265 @@
+/**
+ * @file Footer
+ * @module Footer
+ * @overview Standard and simplified site footer
+ */
+
+.footer {
+  color: color(background);
+  background-color: color(primary);
+  overflow: hidden; // Remove space below footer in IE 10/11
+}
+
+.footer-top {
+  padding: ms(5) ms(4) ms(1);
+  background-color: color(primary);
+
+  @include respond-between($break-mobile, $break-medium) {
+    padding: ms(5) ms(5) ms(1);
+  }
+
+  @include respond-between($break-medium, 1100px) {
+    padding: ms(6) ms(6) ms(1);
+  }
+
+  @include respond-at-and-above(1100px) {
+    padding: ms(8) ms(6) ms(4);
+  }
+
+  @include respond-at-and-above(1178px) {
+    padding-left: 0;
+    padding-right: 0;
+  }
+}
+
+.footer-bottom {
+  background-color: color(dark-blue);
+  padding: 0 ms(5) ms(-2);
+
+  @include respond-at-and-above($break-medium) {
+    padding-bottom: ms(5);
+  }
+}
+
+.footer-nav {
+  position: relative;
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.footer-nav-left {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  justify-content: flex-start;
+  padding-bottom: ms(4);
+}
+
+.footer-nav-group {
+  padding-right: ms(5);
+  display: inline-block;
+  vertical-align: top;
+
+  &.footer-nav-group--bottom {
+    padding-right: 0;
+  }
+}
+
+.footer-nav-group--right {
+  padding-right: 0;
+  margin-left: auto;
+}
+
+.footer-nav-group--links {
+  padding-right: ms(5);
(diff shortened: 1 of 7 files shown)


src/footer/_base.scss
```diff
@@ -0,0 +1,265 @@
...
+.footer-address {
+  display: none;
+
+  @include respond-at-and-above(1080px) {
+    display: inline-block;
+  }
+}
+
+// Simplified Footer
+.footer--simple {
```
misscs: For the sake of keeping files as single purpose as possible, let's move the simplified footer into its own partial

src/footer/footer.macros.html
```diff
@@ -0,0 +1,59 @@
+{#
+  * @overview Macros for standard and simplified site Footers
+#}
+
+{% from '../icons/icons.macros.html' import icon as icon %}
+
+{#
+  * Renders a standard or simplified footer
+#}
+{% macro footer(simple=false, class=null, id=null) %}
```
misscs, line 10: I'd rather use `true`, `false`, and `null` consistently. What do you think of standardizing on `true` and `false`?

src/footer/footer.macros.html
```diff
@@ -0,0 +1,59 @@
...
+<footer class="footer{% if simple %} footer--simple{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
+  <div class="footer-top r-inner">
+    {{ caller() }}
+  </div>
+  <div class="footer-bottom">
+    <div class="footer-nav r-inner">
```
misscs: Let's either use the `nav` element or `role="navigation"`

src/footer/footer.macros.html
```diff
@@ -0,0 +1,59 @@
...
+      <div class="footer-nav-group footer-nav-group--bottom">
+        <span class="footer-nav-country-icons">
+          <a class="footer-nav-group-link footer-nav-group-link--country" href="?country_id=49">
+            {{ icon('usa', size='small') }}
+          </a>
+          <a class="footer-nav-group-link footer-nav-group-link--country" href="?country_id=204">
+            {{ icon('canada', size='small') }}
+          </a>
+        </span>
+        <div class="footer-nav-group-text--mobile">
```
misscs, line 26: Our helper classes for hiding on different screen sizes would work well here. 

Overall, I think we can and should simplify this footer. The classes are unwieldy. Eddie's PR has already been merged. Let's tackle our cleanup and refactor for the simple version. We can port over our updates to Casper.

src/footer/footer.macros.html
```diff
@@ -0,0 +1,59 @@
...
+
+{#
+  * Renders a standard or simplified footer
+#}
+{% macro footer(simple=false, class=null, id=null) %}
+<footer class="footer{% if simple %} footer--simple{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
+  <div class="footer-top r-inner">
+    {{ caller() }}
+  </div>
+  <div class="footer-bottom">
```
misscs, line 15: To help clarify that we are referring to regions of the footer what do you think of either `footer-region-top` or `r-footer-top` where `r` continues as our `region` namespace. 

This change would apply to L15 and L12

misscs:
@josephco Reviewed. It would be good to put in a simple footer that we're happy about. Unfortunately Eddie's footer PR is in without as much refactoring as I would have hoped. I think that's fine because we can work on the footer as a series of patterns, the main one being a stacked list of links with a header. For now, let's get the simple portion feeling like code we'd want to work with.

src/footer/footer.macros.html
josephco, line 10: @misscs do you mean to use false instead of null? i've only been using true/false for booleans

src/footer/footer.macros.html
josephco: @misscs Agree that the footer needs simplification. I'm gonna have to come back to it when I have more time after the landing fit andfinish.

src/footer/footer.macros.html
josephco, line 15: @misscs If we are happy keeping the r- namespace for regions I don't mind going with that, although it feels sorta separate to the footer module when prefixed

josephco:
@misscs I've made the easier fixes to this, but would like to come back to cleaning up the simple footer when I have more time. Could we merge this as is, and have a cleanup task?

PR #56, opened 2016-03-23
Create Backgrounds module and Starry Background JS

josephco:
This is tied to CasperSleep/nightshade#59, which should be merged after this
This should be merged before #53


src/backgrounds/_base.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+h1,
+h2,
+h3 {
+  .background--primary &,
+  .background--starry & {
+    color: color(background);
+  }
+}
```
misscs: Hey, @ipangandoyon! What's the best way for us to note requests for assets? This request is actually a long standing one to provide the starry night background as an SVG. Our current implementation is not very performant and unwieldy from a code maintenance perspective. 

This ask isn't urgent. It would be great if it could happen as part of the month iteration as it can be seen as part of what's necessary to release this pattern. However, it isn't necessary as long as we can get it tee'd up.

misscs: @josephco L106-112, can we set color on the parent class to `color(background)`? If not, although it would be more typing, it would be cleaner to merge these styles into the parent class ruleset, i.e. 

``` sass
.background--primary {
 h1,
 h2, 
 h3 { // styles }
}
```

src/backgrounds/_base.scss
```diff
@@ -0,0 +1,113 @@
+/**
+ * @file Backgrounds
+ * @module Backgrounds
+ * @overview Background and matching text styles for page modules
+ * @todo: remove _backdrops.scss since this replaces it?
+ * @todo: rename TrialWarranty to be generic, update to use Starry Background
+ */
+
+.background--gray {
```
misscs: It's hard to do, but it is best to avoid creating super specific styling classes. Can we use `light` or a different key? `base`? `default`? `background--muted`? Would be great to hear what design calls it internally. Maybe we can glean a good class name from their usage. Unless they just call it gray :stuck_out_tongue_winking_eye:

src/backgrounds/_base.scss
```diff
@@ -0,0 +1,113 @@
+/**
+ * @file Backgrounds
+ * @module Backgrounds
+ * @overview Background and matching text styles for page modules
+ * @todo: remove _backdrops.scss since this replaces it?
```
misscs, line 5: Yes, let's remove `backdrops`.

src/backgrounds/_base.scss
```diff
@@ -0,0 +1,113 @@
...
+ * @todo: rename TrialWarranty to be generic, update to use Starry Background
```
misscs, line 6: :thought_balloon: I was thinking about Trial Warranty on Monday after reviewing the callout pattern. It's basically a callout. Only difference is that it is defined. In fact we have other defined callouts, another example that comes to mind is the newsletter signup half callout/block. 

I'm starting to think that it may behoove us to avoid having too many defined modules in core. I would prefer to focus on providing generic layouts, modules, and element combinations that we can use to compose larger patterns. 

I raise this thought because I wonder if the concept of the Trial Warranty as its own pattern is one we want to keep. I can see it being its own pattern if it always has certain unique styling, say for the icon at different breakpoints, but if it just composes itself off existing patterns with little unique CSS, maybe we don't need to call it out.

</thoughts>

src/backgrounds/StarryBackground.js
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+export const StarryBackground = {
+
+  /**
+   * Finds page modules with the Starry Background selector and triggers setup
+   * @returns {void}
+  */
+  init() {
+    this.StarryBackgroundEls = document.querySelectorAll(`.background--starry`);
```
misscs, line 13: Should this be a JS class or is this an exception? If this an exception, then what's the rule...is it because the JS is really just applying styles? 

Or could we use data attributes? What do we think of that alternative `data-background="starry"? Would we extend the idea of data attributes for backgrounds to all our background patterns?

src/backgrounds/_base.scss
josephco: I'll use `light` for this, and discuss with Becki when she's back, as I'm just porting what was in backdrops, and not clear on whether there are other styles I'm excluding.

src/backgrounds/_base.scss
josephco, line 6: I do think this kinda removes the need for TrialWarranty being its own thing, since it's pretty much a callout, but if it's easier for people to just drop it in with the right styling tweaks, maybe it can live

src/backgrounds/StarryBackground.js
josephco, line 13: I figured since it's a module that's being defined in our core components, it is being authored explicitly with certain classnames. The `.background--starry` class name is the foundation necessary for the background to even show up, but I guess it wouldn't be harmful to add a separate JS class to be strict about our conventions if that's preferred

I don't mind the idea of using data-attributes, but then we'd be selecting in JS by `document.querySelectorAll('[data-background=""]')`, which is fine, too I guess. My instinct is that the data-attr way is better for defining components like buttons and icons, rather than adding style to an existing component

src/backgrounds/_base.scss
josephco, line 5: Removed

PR #57, opened 2016-03-24
Add basic Gallery module using Flickity

josephco:
This is tied to https://github.com/CasperSleep/nightshade/pull/60, which should be merged after this

### DIFF
--- a/src/gallery/_base.scss
+++ b/src/gallery/_base.scss
@@ -1,97 +1,63 @@
+@import "https://cdnjs.cloudflare.com/ajax/libs/flickity/1.2.1/flickity.min.css";
+
 .gallery {
   position: relative;
   overflow: hidden;
+  max-height: ms(16);
 
   &:hover {
-    .gallery-controls {
+
+    .flickity-prev-next-button {
       opacity: 1;
     }
   }
-}
 
-.gallery-slider {
-  width: 100%;
-  height: 100%;
-  white-space: nowrap;
-}
-
-// @@@ [1] inflexible positioning and size should not be determined
-// by the component. let each instance define their own or set
-// overideable defaults
-.gallery-controls {
-  @include z-index(gallery-controls);
-
-  @include square(65px);
-
-  border-radius: 33px;
-  transition: opacity 0.2s ease, background-color 0.15s;
-  position: absolute;
-  top: 50%;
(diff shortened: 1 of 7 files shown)


src/gallery/_gallery.scss
```diff
@@ -0,0 +1,97 @@
...
+  transition: opacity 0.2s ease;
+  margin: 6px;
+  display: inline-block;
+  background-color: rgba(0, 0, 0, 0.4);
+  background-size: contain;
+}
+
+.gallery-slide-wrapper {
+  @include square(100%);
+}
```
misscs, line 97: Are we still using this file?

src/gallery/_base.scss
```diff
@@ -1,97 +1,63 @@
+@import "https://cdnjs.cloudflare.com/ajax/libs/flickity/1.2.1/flickity.min.css";
+
```
misscs: The problem with this approach is Sass treats it as a CSS import so we take a hit with an extra HTTP request. For now, let's move a copy of the css file into our `nightshade-core` pattern folder and we can import it as we do any local css/sass file.

misscs:
:+1: One small change otherwise DTM. 

I added another card in our backlog https://trello.com/c/PkQYzMhp. Do you agree with the problem and intent or do you not think it is that big of a pain point?

misscs:
When you're done making the change, you can give it DTM label.

src/gallery/_gallery.scss
josephco, line 97: @misscs I think it's being used on the storefront static pages, but I'm gonna strip all this out later when I use the new gallery module on those (once it's working with images).

src/gallery/_base.scss
josephco: @misscs ah, good catch. I was thinking about pulling in the styles we want and adding it to the stylesheet, but for now I'll pull the file in locally and include. I think eventually we can restyle the Flickity controls ourselves, giving design an opportunity to think about them and just having our own CSS for it.

josephco:
@misscs I agree that it may be confusing if `nightshade-core` modules rely on dependencies installed elsewhere, since we want modules to be self sufficient. Our core patterns should probably only rely on what's provided by Ando. It's something we will have to think about when we get to that card.

PR #58, opened 2016-03-25
Module improvements

josephco:
This is tied to https://github.com/CasperSleep/nightshade/pull/61, which should be merged after this

### DIFF
--- a/src/animation/ScrollTo.js
+++ b/src/animation/ScrollTo.js
@@ -15,7 +15,7 @@ export const ScrollTo = {
   scroll({scrollTarget, offset, speed}) {
     this.offset = offset || 0;
     this.targetPosition = scrollTarget.offsetTop - this.offset;
-    this.scrollPosition = window.scrollY;
+    this.scrollPosition = window.pageYOffset;
     this.speed = speed || 1000;
     this.time = Math.max(0.1, Math.min(Math.abs(this.scrollPosition - this.targetPosition) / this.speed, 0.8));
     this.currentTime = 0;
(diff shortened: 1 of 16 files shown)


src/content_panel/panel.macros.html
```diff
@@ -5,10 +5,10 @@
 {#
   * Renders a content and image panel.
 #}
-{% macro content_panel(heading, image_path, image_description=null, image_position='left', class=null, id=null) %}
+{% macro content_panel(heading, image_path, image_description=null, image_position='left', imgix_class=null, class=null, id=null) %}
```
misscs: I dig `imgix_class` as the standard param key for any module that has media coming from imgix.

src/animation/ScrollTo.js
```diff
@@ -0,0 +1,57 @@
...
+    const ease = this._ease(position);
+
+    if (position < 1) {
+      window.requestAnimationFrame(this._tick.bind(this));
+      window.scrollTo(0, this.scrollPosition + ((this.targetPosition - this.scrollPosition) * ease));
+    } else {
+      window.scrollTo(0, this.targetPosition);
+    }
+  }
+};
```
misscs, line 57: This is a tiny enough module to refactor to provide a closure for the methods. Then they'll be "private" :mag: 

```
export const modulename = {

// private method here, no need to use underscore prefix

// another private

return {
 // return scroll()
}


};

```

josephco, line 57: I tried doing it this way, but ran into some JS errors. It seems it has to be defined as a self-invoking function and rethought somewhat. I'd like to come back to this after talking with Rob.

PR #64, opened 2016-04-05
Tooltips and popovers updates

josephco:
Tied to https://github.com/CasperSleep/nightshade/pull/72, which should be merged after this

### DIFF
--- a/src/popovers/_base.scss
+++ b/src/popovers/_base.scss
@@ -61,7 +61,7 @@
     transform: translateX(-50%);
 
     &:after {
-      @include arrow('bottom', background, 15px, $shadow: true, $shadow-opacity: 0.05);
+      @include arrow('bottom', $arrow-color, $arrow-size, $shadow: true, $shadow-opacity: $arrow-shadow-opacity);
     }
 
     .no-touch &[aria-expanded="true"] {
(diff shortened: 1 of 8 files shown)


src/utilities/_arrows_helpers.scss
```diff
@@ -16,6 +16,7 @@
 /// when using arrows for tooltips or tabs. `false` for pure shapes.
 /// @param {bool} shadow [false] - Should the arrow have a shadow. Often
 /// `true` for tooltips
+/// @param {value} shadow-opacity - Opacity for arrow shadows
```
misscs: Don't forget to document the default value

src/popovers/_base.scss
```diff
@@ -103,7 +103,7 @@
     transform: translateY(-50%);
 
     &:after {
-      @include arrow('right', background, 15px, $shadow: true);
+      @include arrow('right', background, 15px, $shadow: true, $shadow-opacity: 0.05);
```
misscs: Given that we are repeating values, I think it would be best to create a `config.scss` for this component and move some of the values into variables. Namely, color (background), arrow-size (15px), and shadow (0.05)

PR #66, opened 2016-04-08
Add purchased Calibre WebFonts

josephco:
Tied to https://github.com/CasperSleep/nightshade/pull/77 and https://github.com/CasperSleep/storefront-static/pull/6 which should be merged after this


src/typography/_fonts_config.scss
```diff
@@ -21,6 +25,16 @@ $serif-stack: Georgia, serif;
  }
 }
 
+%calibre-medium {
+ font-family: $sans-stack;
+ font-weight: 400;
+ font-style: normal;
```
misscs, line 31: Do we need `font-style` property?

josephco, line 31: @misscs I don't know tbh, we have been doing it as a convention for all of the font extends... seems likely that it's unneccessary

PR #68, opened 2016-04-08
Faqs integration branch

ckeisato:
Tied to
https://github.com/CasperSleep/nightshade/pull/79

### DIFF
--- a/src/forms/_input_group.scss
+++ b/src/forms/_input_group.scss
@@ -1,0 +1,63 @@
+// Input group has an input and a button_one
+// If the button has text, it is alongside the input field, if it has an icon, it is
+// positioned absolutely on top
+
+.input-group {
+  display: flex;
+
+  .input {
+    @extend %body--sans-small;
+
+    width: 100%;
+    border: 1px solid color(border);
+    margin: 0;
+    padding: size(smaller) size(small);
+    color: color(dark-text);
+    border-radius: size(normal-corners);
+    background: color(background);
+    cursor: auto;
+
+    &:focus {
+      border-color: color(gray--light3);
+    }
+  }
+
+  [data-button] {
+    padding: size(smaller) size(small);
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+}
+
+.input-group--button {
+  .input {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+
+  .input,
+  [data-button] {
+    height: 48px;
+  }
+}
+
+.input-group--button-icon {
+  position: relative;
+
+  .input {
+    padding-right: 60px;
+    height: 48px;
+  }
+
+  [data-button] {
+    position: absolute;
+    right: 1px;
+    top: 1px;
+    height: 46px;
+    border: 0;
+  }
+
+  [data-icon] {
+    @include square(16px);
+  }
+}
--- a/src/forms/_input_group.scss
+++ b/src/forms/_input_group.scss
@@ -1,6 +1,9 @@
-// Input group has an input and a button_one
-// If the button has text, it is alongside the input field, if it has an icon, it is
-// positioned absolutely on top
+/**
+ * @file Input Group
+ * @module Input Group
+ * @overview Input group has an input and a button.  If the button has text, it
+ * is alongside the input field, if it has an icon, it is positioned absolutely on top
+ */
 
 .input-group {
   display: flex;
--- a/src/forms/forms.macros.html
+++ b/src/forms/forms.macros.html
@@ -24,3 +24,21 @@
 </form>
 <div class="email-signup-response-text js-email-signup-response-text"></div>
 {% endmacro %}
+
+
+{#
+  * Renders an input group of an input and button.
+  * @param {string} input_name - Name attribute for the input element
+  * @param {string} [input_placeholder] - Placeholder text for the input
+  * @param {string} [button_content=''] - Button text content
+  * @param {string} [button_style=primary] - Button style
+  * @param {string} [button_icon] - Icon name for Buttons with icons
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+{% macro input_button(input_name, input_placeholder=null, button_content='', button_style='primary', button_icon=null, class=null, id=null) %}
+<div class="input-group {% if button_icon %}input-group--button-icon{% else %}input-group--button{% endif %}{% if class %} {{ class }}{% endif %}" {% if id %} id="{{ id }}"{% endif %}>
+  <input class="input" name="{{ input_name }}" placeholder="{{ input_placeholder }}" required>
+  {{ button(button_content, style=button_style, icon=button_icon, type="submit") }}
+</div>
+{% endmacro %}
--- a/src/forms/forms.macros.html
+++ b/src/forms/forms.macros.html
@@ -1,5 +1,6 @@
 {#
   * @overview Collection of macros for rendering form elements.
+  * @TODO Keep generic form and button macro but make specific ones that build on top, ex: search
 #}
 
 {% from 'nightshade-core/src/buttons/buttons.macros.html' import button %}
--- a/src/forms/forms.macros.html
+++ b/src/forms/forms.macros.html
@@ -33,12 +33,13 @@
   * @param {string} [button_content=''] - Button text content
   * @param {string} [button_style=primary] - Button style
   * @param {string} [button_icon] - Icon name for Buttons with icons
+  * @param {boolean} [required=true] - Boolean for whether or not the input is required
   * @param {string} [class] - Class name for style or JS targeting
   * @param {string} [id] - ID for JS targeting
 #}
-{% macro input_button(input_name, input_placeholder=null, button_content='', button_style='primary', button_icon=null, class=null, id=null) %}
+{% macro input_button(input_name, input_placeholder=null, button_content='', button_style='primary', button_icon=null, required='true', class=null, id=null) %}
 <div class="input-group {% if button_icon %}input-group--button-icon{% else %}input-group--button{% endif %}{% if class %} {{ class }}{% endif %}" {% if id %} id="{{ id }}"{% endif %}>
-  <input class="input" name="{{ input_name }}" placeholder="{{ input_placeholder }}" required>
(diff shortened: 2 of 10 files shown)


src/hero/hero.macros.html
```diff
@@ -12,14 +12,14 @@
...
 {% macro hero(heading, content=null, background_path=null, featured=false, imgix_class='imgix-fluid--nocss', class=null, id=null) %}
 <header class="hero{% if featured %} hero--featured{% endif %} {{ imgix_class }}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}{% if background_path %} data-src="https://casper.imgix.net/{{ background_path }}"{% endif %}>
   <h1 class="hero-heading">{{ heading }}</h1>
   {% if content %}
   <h2 class="hero-subheading">{{ content }}</h2>
+  {% else %}
+  {{ caller() }}
   {% endif %}
 </header>
 {% endmacro %}
```
misscs, line 25: @ckeisato / @josephco: 

This is a good time to set some standards for readability. Personally, I find the macro hard to parse without some whitespace and indentation. Here are some proposals, what do ya think? :) 

Option a: No whitespace, indent in control loops 

```
<header class="hero{% if featured %} hero--featured{% endif %} {{ imgix_class }}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}{% if background_path %} data-src="https://casper.imgix.net/{{ background_path }}"{% endif %}>
  <h1 class="hero-heading">{{ heading }}</h1>
  {% if content %}
    <h2 class="hero-subheading">{{ content }}</h2>
  {% else %}
    {{ caller() }}
  {% endif %}
</header>
```

Option b: Indent and whitespace

```
<header class="hero{% if featured %} hero--featured{% endif %} {{ imgix_class }}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}{% if background_path %} data-src="https://casper.imgix.net/{{ background_path }}"{% endif %}>
  <h1 class="hero-heading">{{ heading }}</h1>

  {% if content %}
    <h2 class="hero-subheading">{{ content }}</h2>
  {% else %}
    {{ caller() }}
  {% endif %}
</header>
```

src/links/links.macros.html
```diff
@@ -0,0 +1,25 @@
...
+
+{#
+  * Renders a link with an icon to the right.
+  * @param {string} content - Link text content
+  * @param {string} href - Path or full URL for link
+  * @param {string} [icon=arrow] - Name of icon that renders and SVG icon
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+{% macro link(content, href, icon='arrow', class=null, id=null) %}
```
misscs: whitespace between comment and macro

src/links/links.macros.html
```diff
@@ -0,0 +1,25 @@
...
+<a class="link{% if icon %} link--icon{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %} href="{{ href }}">
+  {{ content }}
+  {% if icon %}
+    {{ icon_macro(icon) }}
```
misscs: what do we think about a default? Will we always use the link macro with certain icon(s)? Weigh in from design here would be great!

src/links/_base.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * @file Link
+ * @module Link
+ * @overview Link and top of page featured link modules
```
misscs: Not sure what page featured link modules means. Perhaps we can clarify this a bit so there is less ambiguity for devs.

src/links/links.macros.html
```diff
@@ -19,7 +19,7 @@
 <a class="link{% if icon %} link--icon{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %} href="{{ href }}">
   {{ content }}
   {% if icon %}
-    {{ icon_macro(icon) }}
+    {{ icon_macro(icon, size="tiny") }}
   {% endif %}
 </a>
 {% endmacro %}
```
misscs, line 36: Not a super big deal in this context, but worth noting so as well work on improving our commits that the update to macro doesn't fit the description of this commit. Perhaps could have been rebased/squashed into the commit the added the macro file. 

Overall though, :100: on small, single purpose commits. <3'ing it

src/links/links.macros.html
```diff
@@ -8,18 +8,29 @@
...
+  * @param {string} content - Link text content
+  * @param {string} href - Path or full URL for link
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+{% macro link(content, href, class=null, id=null) %}
+<a class="link{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %} href="{{ href }}">
+  {{ content }}
+</a>
+{% endmacro %}
```
misscs, line 21: I've found from past projects that having a granular macro such as link often created more headache than writing the code. We didn't have content in yaml though so that could be the problem. 

It is fine to keep this as sometimes using the macro make keep code more readable!

src/lists/lists.macros.html
```diff
@@ -0,0 +1,34 @@
...
+    </li>
+    {% else %}
+    <li>
+      {{ link(item.content, href=item.href) }}
+    </li>
+    {% endif %}
+  {% endfor %}
+  </ul>
+</div>
+{% endmacro %}
```
misscs, line 36: Since this pattern specifically is a list with a heading, being more explicit in our naming would make it clearer what this does. I would assume a list macro just makes a list.

src/lists/_base.scss
```diff
@@ -5,5 +5,5 @@
  */
 
 .list {
-  padding: ms(1) 0;
+  margin: ms(1) 0;
```
misscs, line 8: @josephco also brought this up, but we may just want to standardize on bottom or top that way we don't need to do any overrides for final elements. Maybe this isn't an issue for this module. Fit and finish could shed some more insight.

src/forms/_input_group.scss
```diff
@@ -0,0 +1,63 @@
+// Input group has an input and a button_one
+// If the button has text, it is alongside the input field, if it has an icon, it is
+// positioned absolutely on top
+
```
misscs, line 7: We'll want to continue the JSDoc style docs at the top of the page. This description is great for @overview directive. One fix would be to just use `button` instead of `button_one`

src/forms/_input_group.scss
```diff
@@ -0,0 +1,63 @@
...
+.input-group {
+  display: flex;
+
+  .input {
+    @extend %body--sans-small;
```
misscs, line 12: Extending this font style feels fragile. curious what @josephco thinks.

src/forms/forms.macros.html
```diff
@@ -24,3 +24,21 @@
...
+  * @param {string} [button_icon] - Icon name for Buttons with icons
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+{% macro input_button(input_name, input_placeholder=null, button_content='', button_style='primary', button_icon=null, class=null, id=null) %}
+<div class="input-group {% if button_icon %}input-group--button-icon{% else %}input-group--button{% endif %}{% if class %} {{ class }}{% endif %}" {% if id %} id="{{ id }}"{% endif %}>
+  <input class="input" name="{{ input_name }}" placeholder="{{ input_placeholder }}" required>
+  {{ button(button_content, style=button_style, icon=button_icon, type="submit") }}
+</div>
+{% endmacro %}
```
misscs, line 46: I find myself going back and forth on this pattern. From one end it feels too abstract and generic, like it accomplishing too many things. But from another end, it feels like it is robust to handle different situations. 

There's a part of me that would prefer to have two explicit patterns: One for search input and the other for an input with a button (newsletter and promo fields on our current site).

src/forms/forms.macros.html
```diff
@@ -24,3 +24,21 @@
...
+  * @param {string} [input_placeholder] - Placeholder text for the input
+  * @param {string} [button_content=''] - Button text content
+  * @param {string} [button_style=primary] - Button style
+  * @param {string} [button_icon] - Icon name for Buttons with icons
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+{% macro input_button(input_name, input_placeholder=null, button_content='', button_style='primary', button_icon=null, class=null, id=null) %}
+<div class="input-group {% if button_icon %}input-group--button-icon{% else %}input-group--button{% endif %}{% if class %} {{ class }}{% endif %}" {% if id %} id="{{ id }}"{% endif %}>
+  <input class="input" name="{{ input_name }}" placeholder="{{ input_placeholder }}" required>
```
misscs: Need to add `type` and make `required` a boolean. Sometimes the input may not be required

src/forms/_input_group.scss
```diff
@@ -0,0 +1,63 @@
...
+  [data-button] {
+    height: 48px;
+  }
+}
+
+.input-group--button-icon {
+  position: relative;
+
+  .input {
+    padding-right: 60px;
```
misscs, line 51: Feels like a magic number!

src/links/links.macros.html
ckeisato, line 21: I'm ok with not having a link macro, would all link styling then be a part of typography?

src/hero/hero.macros.html
ckeisato, line 25: Do you want new lines for control loops or between elements?  Or both?  I'm fine either way but if we're going to have more whitespace in the macro, I think @josephco was saying we have to strip the output (which we should probably do anyway).

src/lists/lists.macros.html
ckeisato, line 36: Sure that makes sense, will be amending this to be `list--heading`.

src/forms/forms.macros.html
ckeisato, line 46: Isn't the search input also an input with a button?  Or do you mean a input with an obvious button (like email sign up) vs an input with a ghost button so that it looks like it's inside the input?

src/lists/lists.macros.html
ckeisato, line 36: @misscs Since we may have lists with or without headings, would it make sense to keep it called `List` but have the heading be optional?

PR #70, opened 2016-04-11
Add list macro and list macro base styling

ckeisato:


src/lists/_base.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * @file List
+ * @module List
+ * @overview List and list with action link modules
+ */
+
+.list {
+  padding: ms(1) 0;
```
josephco: @ckeisato I've been using margin-bottom as a convention for spacing things relative to other modules etc. And padding more for component internal padding etc. The styling for this module will probably change, but it might be a good convention to follow.

src/lists/lists.macros.html
```diff
@@ -0,0 +1,34 @@
+{#
+  * @overview Macros for Lists.
+  * @module Lists
+  *
+#}
+
+{% from 'nightshade-core/src/links/links.macros.html' import link, link_icon %}
+
+{#
+  * Renders a list of links
```
josephco: I'm following a convention to always refer to modules in title case, so any mention of the List module in the documentation should be capitalized. If y'all think it's overkill we can revise, but for now let's do that.

src/lists/_base.scss
ckeisato: Yeah sounds good, amending this now.

josephco:
@ckeisato Two small comments, otherwise DTM

PR #71, opened 2016-04-11
Cleanup core imports

josephco:
Refactor core imports to stop generating output and group related styles together

### DIFF
--- a/src/layout/_base.scss
+++ b/src/layout/_base.scss
@@ -46,3 +46,8 @@ body {
     box-sizing: inherit;
   }
 }
+
+// Disable 300ms delay on touch devices
+.no-touch-delay {
+  touch-action: manipulation;
+}
--- a/src/layout/_helper_classes.scss
+++ b/src/layout/_helper_classes.scss
@@ -1,81 +1,38 @@
-// Layout
-//
-// 1. Spree-auth-devise hax which I hate
-//
-// @@@ TODO
-// This file should just be positioning/layout helper classes
+// Layout helper classes
 
-// class for defining sections of content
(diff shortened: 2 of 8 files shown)


src/layout/_base.scss
```diff
@@ -46,3 +46,8 @@ body {
     box-sizing: inherit;
   }
 }
+
+// Disable 300ms delay on touch devices
+.no-touch-delay {
+  touch-action: manipulation;
+}
```
misscs, line 53: Ooo, like this location for this code. +1

src/layout/_helper_classes.scss
```diff
@@ -1,81 +1,38 @@
...
   margin-left: auto;
   margin-right: auto;
 }
 
-
 .-hoisted {
   @include respond-at-and-above(1200px) {
     margin-top: -40px;
   }
 }
```
misscs, line 43: Let's go ahead and remove lines 29-38. It is old code and we don't have these concepts in core yet. may as well keep this lean.

src/layout/_helper_classes.scss
```diff
@@ -1,81 +1,38 @@
-// Layout
-//
-// 1. Spree-auth-devise hax which I hate
-//
-// @@@ TODO
-// This file should just be positioning/layout helper classes
+// Layout helper classes
```
misscs: Please add file level docs :)

src/layout/_helper_classes.scss
josephco, line 43: Added a todo to clean this up when we get to static storefront stuff

PR #72, opened 2016-04-12
Style Banners with classes, not attr selectors

josephco:
Tied to https://github.com/CasperSleep/nightshade/pull/83, which should be merged after this

### DIFF
--- a/src/banners/_base.scss
+++ b/src/banners/_base.scss
@@ -1,7 +1,7 @@
 // short full width product banner, on all products
 // currently has free shipping and returns message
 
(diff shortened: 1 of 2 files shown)


src/banners/_base.scss
```diff
@@ -22,6 +22,6 @@
   }
 }
 
-[data-banner="background"] {
+.banner--background {
```
misscs, line 25: I'm not married to this class. It doesn't feel quite clear enough. Would rather let the team decide on this though.

PR #74, opened 2016-04-12
Remove color macros (moved to Nightshade)

josephco:
Tied to https://github.com/CasperSleep/nightshade/pull/87, which should be merged before this.


misscs:
There's a very tight coupling on this PR which makes it hard to merge. It is always better to err on more, smaller PRs. We can't merge the color branch as is because it is WIP. 

Can you create a smaller PR on nightshade that just deals with adding this macro so we can close this PR?

josephco:
@misscs Sure will do that, it needs to come into nightshade either way

misscs:
Cool https://github.com/CasperSleep/nightshade/pull/89 looks good too

PR #75, opened 2016-04-14
Add accordion module

ckeisato:


src/accordion/Accordion.js
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+export const Accordion = {
+
+  /**
+   * Inits accordion elements
+   * @returns {void}
+  */
+  init() {
+    this.accordionSection  = document.getElementsByClassName(`js-accordion-section`);
```
josephco: Do you think we should just use .querySelector/.querySelectorAll to standardize to be more like the $() in jQuery? .getElementById seems right for ID targeting, otherwise I think querySelector is more modern and practical.

src/accordion/Accordion.js
```diff
@@ -0,0 +1,55 @@
...
+
+    for (let i = 0; i < this.accordionSection.length; i++){
```
josephco: We can use the ES6 spread [...] thing here, see how it's done in Tooltips/Popovers.js

src/accordion/Accordion.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+      });
+    }
+  },
+
+  /**
+   * Opens the given Accordion section
+   * @returns {void}
+  */
+  openAccordionSection(section){
```
josephco: Space before the { and on line 51

src/accordion/Accordion.js
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Inits accordion elements
+   * @returns {void}
+  */
+  init() {
+    this.accordionSections  = document.querySelectorAll(this.accordionSection);
+
+    [...this.accordionSections].forEach((el) => {
+      let trigger = el.querySelectorAll(this.accordionSectionTrigger);
```
josephco: @ckeisato You could just use `el.querySelector(this.accordionSectionTrigger);` instead of having to do line 22

src/accordion/Accordion.js
```diff
@@ -0,0 +1,59 @@
...
+      trigger = trigger[0];
+      let content = el.querySelectorAll(this.accordionSectionContent);
```
josephco: Same for this

src/accordion/Accordion.js
```diff
@@ -0,0 +1,59 @@
...
+      content = content[0];
+
+      trigger.setAttribute(`touch-action`, `none`);
+      content.setAttribute(`aria-expanded`, false)
+
+      trigger.addEventListener('pointerup', (e) => {
```
josephco: Need backticks for strings

src/accordion/Accordion.js
```diff
@@ -0,0 +1,59 @@
...
+        e.stopPropagation();
+
+        if (content.classList.contains(this.openClass)) {
+          this.closeAccordionSection(content);
+        }
```
josephco: Do we have a convention to have else on the same line as closing if line? eg `} else {`

src/accordion/Accordion.js
```diff
@@ -0,0 +1,59 @@
...
+  accordionSectionTrigger: `.js-accordion-section-toggle`,
+  accordionSectionContent: `.js-accordion-section-content`,
+  openClass: `is-open`,
+
+  /**
+   * Inits accordion elements
+   * @returns {void}
+  */
+  init() {
+    this.accordionSections  = document.querySelectorAll(this.accordionSection);
```
josephco: This could just be a const if it's not being used outside of init(), then you won't have to use `this.`

src/accordion/accordion.macros.html
```diff
@@ -0,0 +1,51 @@
+{#
+  * @overview Macros for Accordions
+  * @module Accordion
+  *
+#}
+
+{% from 'nightshade-core/src/icons/icons.macros.html' import icon, icon %}
```
josephco: Icon is being imported twice

PR #79, opened 2016-04-18
Update Hero macro to use picture element

josephco:


misscs:
Closing this PR. We ended up staying with imgix.js. After talking with Imgix team, moving to picturefill and srcset is now recommended over imgixjs.

PR #80, opened 2016-04-19
Add accordion module

ckeisato:


src/accordion/Accordion.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * @overview An Accordion module comprised of sections that toggle open and close
+ * @module Accordion.js
+*/
+
+export const Accordion = {
+
+  accordionID: ``,
```
misscs: What if it is a class? ID feels very specific. Maybe something like accordingEl?

src/accordion/_base.scss
```diff
@@ -0,0 +1,42 @@
...
+   * @module Accordion
+   * @overview
+   * @@@ todo: standardize height transition, also make proportional to max-height/media queries
+ **/
+
+.accordion-list-item {
+   margin: ms(1) 0;
+ }
+
+.accordion-section {
```
misscs: Curious to why we need to nest classes in this instance?

src/accordion/_base.scss
```diff
@@ -0,0 +1,42 @@
...
+    cursor: pointer;
+
+    [data-icon="arrow"] {
+      transform: rotate(90deg);
+    }
+  }
+
+  &.is-open {
+    .accordion-section-list {
+      max-height: 1000px;
```
misscs: This feels dangerous. Why do we need to set max-height?

src/accordion/_base.scss
```diff
@@ -0,0 +1,42 @@
...
+
+      @include respond-at-and-above($break-medium) {
+        max-height: 500px;
```
misscs: Same comment here. Where is this value coming from?

src/accordion/_base.scss
ckeisato: @misscs This isn't ideal, but we don't have a good way to animate height at the moment.  These values would also have to be reset depending on the content for each accordion.

src/accordion/Accordion.js
ckeisato: This is amended.

src/accordion/_base.scss
ckeisato: This is amended.

PR #84, opened 2016-04-20
Add breadcrumbs module

ckeisato:


src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -1,17 +1,37 @@
...
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+
+{% macro breadcrumbs(data, color=null, class=null, id=null) %}
+<div class="breadcrumbs {{ color }}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
+  <ol class="breadcrumbs-list">
+    <li class="breadcrumbs-list-item breadcrumbs-logo">
+      <a class="" href="#">
+        {{ icon("casper-logo-mark", size="small" )}}
```
josephco: @ckeisato let's stick to single quotes for declaring macro properties

src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -1,17 +1,37 @@
...
+{#
+  * Renders Breadcrumbs
+  * @param {array} data - List of links for Breadcrumbs
+  * @param {string} [color=light] - Indicates the color theme for the Breadcrumbs
+  * @param {string} [class] - Class name for style or JS targeting
+  * @param {string} [id] - ID for JS targeting
+#}
+
+{% macro breadcrumbs(data, color=null, class=null, id=null) %}
+<div class="breadcrumbs {{ color }}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
```
josephco: @ckeisato The param doc says default is 'light', but no default is being set. Would be good to use BEM syntax: `breadcrumbs--light`, and this will help with organization of the sass too

src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -1,17 +1,37 @@
...
+  <ol class="breadcrumbs-list">
+    <li class="breadcrumbs-list-item breadcrumbs-logo">
```
josephco: not sure we should be applying the logo class to the list item row. it's not really following BEM structure, is this class being used?

src/breadcrumbs/breadcrumbs.macros.html
ckeisato: Sorry about the param, that was left over.  Isn't it weird to pass in the class `breadcrumbs--light` as the `color` param?
So you'd call the macro as
`{{ breadcrumbs(mattress_data, color='breadcrumbs--light') }}`

ckeisato: Or should I append the color to a class in the macro?

src/breadcrumbs/breadcrumbs.macros.html
ckeisato: Amended

src/breadcrumbs/breadcrumbs.macros.html
ckeisato: Amended

src/breadcrumbs/breadcrumbs.macros.html
ckeisato: Yep, I'm appending the color argument to a bem class

src/breadcrumbs/_base.scss
```diff
@@ -2,60 +2,40 @@
...
  */
- 
-.breadcrumb-list {
-  padding-bottom: ms(3);
-  list-style: none;
 
-  a {
-    display: inline-block;
+.breadcrumbs {
+  .breadcrumbs-link {
```
josephco, line 10: We shoud use a linespace inside blocks (and on line 24)

ckeisato, line 10: amended

PR #88, opened 2016-04-22
Add more base styling to accordion pattern

ckeisato:
This PR adds some more base styling to the accordion pattern
1)  Bottom border on all but the last accordion section
2)  Adjust spacing within the accordion toggle to increase target area
3)  Use flexbox to style the accordion toggle content and icon
4)  Use velocity to open and close the accordion
5)  Create a preliminary ie stylesheet, this will need more testing

I'm not sure why the icons aren't showing up in the nightshade repo at the moment, but the little black square is where it would be.

screen shot 2016-04-22 at 4 31 48 pm

### DIFF
--- a/src/accordion/Accordion.js
+++ b/src/accordion/Accordion.js
@@ -13,6 +13,7 @@ export const Accordion = {
   /**
    * Inits accordion elements
    * @param {string} accordionEl Accordion DOM Element
+   * @param {integer} [time=300] Duration for open/close transition in the Accordion
    * @returns {void}
   */
   init(accordionEl, time = 300) {
--- a/src/accordion/Accordion.js
+++ b/src/accordion/Accordion.js
@@ -6,6 +6,7 @@
 export const Accordion = {
 
   accordionSection : `.js-accordion-section`,
+  accordionSectionContent : `.js-accordion-section-content`,
   accordionSectionTrigger: `.js-accordion-section-toggle`,
   openClass: `is-accordion-open`,
 
@@ -27,10 +28,12 @@ export const Accordion = {
       trigger.addEventListener(`pointerup`, (e) => {
         e.stopPropagation();
 
+        let sectionContent = el.querySelectorAll(this.accordionSectionContent);
+
         if (el.classList.contains(this.openClass)) {
-          this.closeAccordionSection(el);
+          this.closeAccordionSection(el, sectionContent);
         } else {
-          this.openAccordionSection(el);
+          this.openAccordionSection(el, sectionContent);
         }
       });
     });
@@ -41,9 +44,10 @@ export const Accordion = {
    * @param {object} section DOM element to toggle state class on
    * @returns {void}
   */
-  openAccordionSection(section) {
+  openAccordionSection(section, sectionContent) {
     section.setAttribute(`aria-expanded`, true)
     section.classList.add(this.openClass);
+    Velocity(sectionContent, `slideDown`);
   },
 
   /**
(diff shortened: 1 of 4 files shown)


src/accordion/_base.scss
```diff
@@ -15,6 +19,7 @@
   transition: max-height 0.3s ease-in-out;
 
   .is-accordion-open & {
+    padding-bottom: ms(1);
     max-height: 1000px;
```
misscs: How is the max value for the accordion being calculated/decided?

src/accordion/_base.scss
```diff
@@ -25,8 +30,11 @@
 
 .accordion-section-toggle {
   cursor: pointer;
+  padding: ms(1) 0;
+  margin-bottom: 0;
 
   [data-icon="arrow"] {
+    float: right;
```
misscs: @beckichoi / @ckeisato: 

The target to open and close the accordion needs to account for variable text length. In its currently implementation, a question has a really long title  will eventually wrap and the alignment with the arrow will be off especially on small devices. 

My recommendation would be to use flexbox to layout the `accordion-section-toggle`—one area of title, one area for the icon. (Design will also want to set rules for length of text to prevent odd text on small devices. Or perhaps we'll want to use ellipsis...There are options for you both to explore during fit and finish!) 

@ArmandoAmador Can you run some queries for us? We are looking for 2-5 examples FAQs with really long titles and a 2-5 examples of short to ave length. At the moment do we limit the length of the title in the CMS?

misscs:
@ckeisato Reviewed the PR. All that's necessary to fix so we can merge is ironing out those two magic numbers and setting up the accordion-toggle-target area with flexbox. The details can wait until FF.

src/accordion/_base.scss
ArmandoAmador: At the moment is the standard 255 character limit set by default, I sent @ckeisato the longest and shortest title example

src/accordion/Accordion.js
```diff
@@ -41,18 +46,20 @@ export const Accordion = {
...
    * @param {object} section DOM element to toggle state class on
    * @returns {void}
   */
-  closeAccordionSection(section) {
+  closeAccordionSection(section, sectionContent) {
     section.setAttribute(`aria-expanded`, false)
     section.classList.remove(this.openClass);
+    Velocity(sectionContent, `slideUp`, this.transitionDuration);
   },
 };
```
misscs, line 65: This is likely another module that we'll want to turn into the View pattern since it will generated on the client. We'll make this call after the search prototype.

src/accordion/_base.scss
```diff
@@ -6,27 +6,31 @@
...
 }
 
 .accordion-section-toggle {
   cursor: pointer;
+  padding: ms(1) 0;
+  margin-bottom: 0;
+  display: flex;
 
   [data-icon="arrow"] {
+    float: right;
```
misscs: The goal with using flex is to remove floats.Try `margin-left: auto` or use one of the flex `align-content`/`align-self` properties to position the arrow.

misscs:
@ckeisato One small fix but otherwise DTM.

src/accordion/Accordion.js
```diff
@@ -27,10 +30,12 @@ export const Accordion = {
       trigger.addEventListener(`pointerup`, (e) => {
         e.stopPropagation();
 
+        let sectionContent = el.querySelectorAll(this.accordionSectionContent);
```
josephco: This may not need .querySelectorAll(), you can probably just use .querySelector() since there's only one el.

PR #90, opened 2016-04-26
Add logo attribute to breadcrumbs

ckeisato:
Adds a boolean attribute to determine if the breadcrumbs should be preceded by the Casper logo


src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -9,27 +9,29 @@
...
+          {{ icon('casper-logo-mark', size='small') }}
+        </a>
       </li>
+    {% endif %}
 
-      <li class="breadcrumbs-list-item">
-        {{ link(item.content, href=item.href, class='breadcrumbs-link') }}
-      </li>
+    {% for item in data %}
+        <li class="breadcrumbs-list-item">
```
josephco: Indented one level too deep

PR #92, opened 2016-04-29
Update breadcrumbs styling

ckeisato:
This PR
1) removes the color option for the breadcrumbs macro
2) replaces the arrow icon with the '>' character
3) removes the use of the link macro in this instance

### DIFF
--- a/src/breadcrumbs/_base.scss
+++ b/src/breadcrumbs/_base.scss
@@ -2,7 +2,6 @@
  * @file Breadcrumbs
  * @module Breadcrumbs
  * @overview Styles for Casper breadcrumbs
- * @TODO: expand/solidify styling, here they have styles to support grey and blue (like on the main site)
  */
 
 .breadcrumbs {
@@ -21,23 +20,33 @@
   }
 }
 
-.breadcrumbs--dark {
-  
+// Styling for Breadcrumbs in the footer
+.breadcrumbs--footer {
+
   .breadcrumbs-link {
-    color: color(blue);
+    color: color(white);
+    opacity: 0.45;
+    transition: 0.25s opacity ease;
+
+    &:hover {
+      opacity: 1;
+    }
   }
 
-  [data-icon="arrow"],
+  .breadcrumbs-list-item {
+    color: color(white);
+    vertical-align: middle;
+
+    &:before {
+      opacity: 0.45;
+    }
+
+    &:first-child {
+      margin-top: -5px;
+    }
+  }
+
   [data-icon="casper-logo-mark"] {
-    fill: color(blue);
-  }
-}
-
-.breadcrumbs-list-item {
-  display: inline-block;
-  margin: 0 ms(-6);
-
-  &:first-child {
-    margin-left: 0;
+    fill: color(white);
   }
 }
--- a/src/breadcrumbs/_base.scss
+++ b/src/breadcrumbs/_base.scss
@@ -4,19 +4,43 @@
  * @overview Styles for Casper breadcrumbs
  */
 
-.breadcrumbs {
+.breadcrumbs-list-item {
+   display: inline-block;
+   margin-left: ms(-7);
+
+   &:before {
+     content: "\003E";
+     margin-right: ms(-7);
+   }
+
+   &:first-child {
+     margin-left: 0;
+
+     &:before {
+       display: none;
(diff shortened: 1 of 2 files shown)


src/breadcrumbs/_base.scss
```diff
@@ -2,40 +2,29 @@
  * @file Breadcrumbs
  * @module breadcrumbs
  * @overview Styles for Casper breadcrumbs
- * @TODO: expand/solidify styling, here they have styles to support grey and blue (like on the main site)
+ * @TODO: create set for footer breadcrumbs
```
misscs: Agreed here. One way to think about this pattern is a breadcrumbs on ligtht background and breadcrumbs on dark background. Maybe something like `breadcrumbs--light-background` and `breadcrumbs--dark-background` but with better modifiers

src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -24,21 +23,25 @@
         </a>
       </li>
       <li class="breadcrumbs-list-item">
-        {{ icon('arrow', size='tiny') }}
+        >
```
misscs: If we're going to use a character, it would be best to use the unicode characters for these symbols to ensure we're okay when we localize.

src/breadcrumbs/_base.scss
ckeisato: That was the original plan, but since @beckichoi said that the footer breadcrumbs would be more of a one off situation, it could make sense to name the `breadcrumbs--dark-background` `breadcrumbs--footer`?  I'd also rather make two separate breadcrumbs macros (footer/logo vs regular) but that seemed like more duplicate code.

src/breadcrumbs/breadcrumbs.macros.html
```diff
@@ -24,21 +23,25 @@
...
-          {{ link(item.content, href=item.href, class='breadcrumbs-link') }}
+          <a class="breadcrumbs-link" href="item.href">
+            {{ item.content }}
+          </a>
         </li>
       {% else %}
         <li class="breadcrumbs-list-item">
-          {{ icon('arrow', size='tiny') }}
+          >
         </li>
```
misscs: Why do we want to wrap the list separator in an `li`, it isn't part of the list. A span within the li or maybe pseudo before/after elements could work here.

src/breadcrumbs/_base.scss
```diff
@@ -29,4 +29,8 @@
   &:first-child {
     margin-left: 0;
   }
+
+  &:last-child {
+    margin-right: 0;
+  }
 }
```
misscs: Is it possible to assign margin on one side so we don't have to zero out both?

src/breadcrumbs/_base.scss
misscs: Is the footer breadcrumb code that unique? Let's sit down together and look at this! I'm still around if you want to do that now. Or tomorrow AM.

src/breadcrumbs/_base.scss
```diff
@@ -24,13 +24,18 @@
 
 .breadcrumbs-list-item {
   display: inline-block;
-  margin: 0 ms(-7);
+  margin-right: ms(-7);
 
-  &:first-child {
-    margin-left: 0;
+  &:after {
+    content: '>';
```
misscs: Don't forget that we want to use unicode here to ensure we're okay in other languages!

src/breadcrumbs/_base.scss
```diff
@@ -16,26 +15,52 @@
...
+    margin-left: 0;
 
-    &:after {
+    &:before {
       display: none;
     }
   }
 }
+
+.breadcrumbs--footer {
```
misscs: Would be nice to have some documentation above each of the different modifiers so that a new person coming to this file understands what they are used for and when

misscs:
I'm DTM pending 
1. Fix to use unicode for `>` 
2. Adding documentation to breadcrumbs sass file

src/breadcrumbs/_base.scss
ckeisato: @misscs What's the translation benefit of using the unicode character over the character itself?
@robwierzbowski was wondering too

misscs:
👍 💃

PR #94, opened 2016-05-03
Adjust accordion bottom spacing for the toggle

ckeisato:


misscs:
Simple and fast. LGTM!

PR #97, opened 2016-05-03
Adjust mobile sides

ckeisato:
This PR increases the side padding above 768px for `.r-section-wrapper`


misscs:
Cool! I'm looking forward to cleaning up this file, but good to see some ms values making their way in!

Issue #100, opened 2016-05-03
Create JS tests for Accordion.js pattern


misscs:
### Vision

Write test for Accordion pattern using testing framework (#96). Achieve file code coverage > 90% (#102)
### Current State

No unit testing currently in `nightshade-core`.
### Scope
- [ ]

stewartyu:
Done

Issue #96, opened 2016-05-03
Setup JavaScript testing framework


stewartyu:
### Vision

Implement unit testing framework for JavaScript widgets. 
### Current State

No unit testing currently in `nightshade-core`.
### Scope
- [x] Implement testing framework (ideally Ava, or Mocha)
### Anti-Vision

Integration/system testing not required at this time.

misscs:
Not going to run it via Gulp because NYC is not available for gulp. The plan is to run tests via NPM script.  Need to decide if we want have npm kick off the gulp process

PR #105, opened 2016-05-04
[96] Unit Testing

stewartyu:
### Issues
#96 - unit testing! This PR is mostly just for comments on general structure and setup of Ava/JSDom.

Started some work on #100 just to have some tests to try out, but will fully flesh that out in a separate PR. 
### Steps to Reproduce

Make sure that `npm run test` and `npm run watch:test` work.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -25,7 +25,7 @@ npm run test
 To continuously run the unit tests as you develop:
 
 ```sh
-npm run watch:test
+npm run test:watch
 ```
 
 ### Labels
--- a/contributing.md
+++ b/contributing.md
@@ -14,6 +14,20 @@ We think it's best for the project if the code you write looks like the code the
 - Add a label to your PR
 - Once your PR is merged, update the version, create a tag, and write a release.
 
+## Testing
+
+To run the unit tests:
+
+```sh
+npm run test
+```
+
+To continuously run the unit tests as you develop:
+
+```sh
+npm run watch:test
+```
+
 ### Labels
 
 We use labels to help tag the state of a PR.
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "test": "ava \"src/**/*/*.test.js\" --verbose",
     "postversion": "git push && git push --tags",
-    "watch:test": "npm run test -- --watch"
+    "test:watch": "npm run test -- --watch"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.0.0",
   "description": "Casper pattern and style scss library. Under development.",
   "scripts": {
-    "test": "ava \"src/**/*/test.js\" --verbose",
+    "test": "ava \"src/**/*/*.test.js\" --verbose",
     "postversion": "git push && git push --tags",
     "watch:test": "npm run test -- --watch"
   },
@@ -23,7 +23,8 @@
     "susy": "^2.2.12"
   },
   "devDependencies": {
-    "ava": "^0.14.0"
+    "ava": "^0.14.0",
+    "jsdom": "^8.5.0"
   },
   "ava": {
     "require": [
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   },
   "devDependencies": {
     "ava": "^0.14.0",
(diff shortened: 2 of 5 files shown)


src/accordion/Accordion.test.js
```diff
@@ -0,0 +1,46 @@
...
+          </div>
+          <ul class="accordion-section-list js-accordion-section-content">
+            <li class="accordion-section-list-item">
+              Content
+            </li>
+          </ul>
+        </div>
+      </li>
+    </ul>
+  </div>`);
```
misscs: It is possible for us use a template so we don't have to maintain markup code in two spots?

contributing.md
```diff
@@ -14,6 +14,20 @@ We think it's best for the project if the code you write looks like the code the
...
+npm run test
+```
+
+To continuously run the unit tests as you develop:
+
+```sh
+npm run watch:test
+```
+
 ### Labels
```
misscs, line 31: 💯 docs!

ArmandoAmador:
This is nice @stewartyu

src/accordion/Accordion.test.js
stewartyu: @misscs Definitely agree! What are we using for those template macros again? I'll see what I can do.

ArmandoAmador:
Why is it called 96 tests for javascript widgets

stewartyu:
@ArmandoAmador That's a great point - it's because it's for issue #96. I should give it a better name though. 😄

misscs:
@stewartyu @ArmandoAmador Maybe we use brackets to denote `[96] Name of Issues`? I do like knowing the issues from the title

package.json
```diff
@@ -3,8 +3,9 @@
   "version": "2.0.0",
   "description": "Casper pattern and style scss library. Under development.",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "postversion": "git push && git push --tags"
+    "test": "ava \"src/**/*/*.test.js\" --verbose",
+    "postversion": "git push && git push --tags",
+    "watch:test": "npm run test -- --watch"
```
robwierzbowski: Why `watch:test` over `test:watch`?

src/accordion/Accordion.test.js
misscs: @stewartyu Currently the precompile task that creates js templates happens in nightshade. One approach, which seems the cleanest,  is to add nunjucks to nightshade-core and a template file to each pattern. Then in tests we just call `nunjucks.render` and pass in the template.

src/accordion/Accordion.test.js
```diff
@@ -0,0 +1,48 @@
...
+            </div>
+            <ul class="accordion-section-list js-accordion-section-content">
+              <li class="accordion-section-list-item">
+                Content
+              </li>
+            </ul>
+          </div>
+        </li>
+      </ul>
+    </div>`);
```
robwierzbowski: Can you import the template file? Copying output adds maintenance overhead and potential for out-of-sync.

robwierzbowski: Ah, I see that C. already noted this.

stewartyu:
Ran into some problems getting Nunjucks templates to render properly, so for now, I'm just using a separate fixture file (`Accordion.test.html`) to split it out of the JS.

src/accordion/Accordion.test.html
```diff
@@ -0,0 +1,29 @@
+`
+<div id="accordion">
```
victorwhy, line 2: is there a reason why we don't just use an existing part of our site to test this? just copy paste an html snippet out of somewhere where we use accordian?

stewartyu, line 2: @victorwhy Good question! Ideal situation is actually to render a Nunjucks template on the fly where we use the `accordion` macro. This way, we'll always be in sync with the latest updates to the HTML structure if the macro ever changes. Also, if the template call resembles a real macro call, then it will be testing exactly the HTML that's being used on production.

The only reason I'm not using the template rendering approach at the moment is because I ran into some issues with Nunjucks file pathing, which Claudina is going to help me with later.

misscs, line 2: Addressing it in #108

misscs:
@stewartyu Tried this out locally. LGTM.

Issue #102, opened 2016-05-03
Setup code coverage report 


misscs:
### Vision

Implement code coverage report (ideally [NYC](https://github.com/bcoe/nyc)) that works with ES6/2015. 
### Current State

No code coverage currently in `nightshade-core`.

PR #106, opened 2016-05-04
[102] Code coverage

stewartyu:
### Issues
#102 code coverage reporting

We don't currently have code coverage reporting, so we're using [https://github.com/bcoe/nyc](NYC) to instrument it in conjunction with our Ava tests.
### Steps to Reproduce

Run `npm run cover` to run it once, or `npm run cover:watch` to continuously run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+.nyc_output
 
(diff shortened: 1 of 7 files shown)


misscs:
DTM. As we start writing tests, I think it will be helpful for folks to have some guidance on how to interpret the coverage table even though we'll rely on our CI to let them know things failed. In short, some docs in frontier-codebook around testing would be great

PR #107, opened 2016-05-05
Add circle.yml to use correct Node version

stewartyu:
### Issues

CircleCI checks are failing on `npm install` because it's using Circle's default Node version (0.10). My hope is that bumping the version to our 5.10.1 will fix this problem.


misscs:
🙏  fingers crossed. thanks!

PR #110, opened 2016-05-05
Update list and link macros

ckeisato:
### Issues

What issues are being solved:
The `link_icon` macro generated a link with an icon at a fixed size.  This PR adds an optional parameter to define the size of that icon.  
Similarly, the list with a cta depended on this macro, so an optional param was added to the list macro as well to pass onto the `link_icon` macro it calls.
### Todos

Add an example to nightshade to show this option.
### Impacted Areas

FAQ
### Steps to Reproduce

On branch v2.0.0-dev you cannot specify the size of the icon for the `link_icon` macro, on this branch you can.

### DIFF
--- a/src/links/links.macros.html
+++ b/src/links/links.macros.html
@@ -24,13 +24,14 @@
   * Renders a link with an icon to the right.
   * @param {string} content - Link text content
   * @param {string} href - Path or full URL for link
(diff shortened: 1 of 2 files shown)


misscs:
💯

PR #112, opened 2016-05-07
Fix breadcrumbs

ckeisato:
### Issues

I recently made some updates to the footer macro, this PR integrates those changes into the regions footer (used throughout Storefront Static).
### Todos

Updating any other footers throughout the site, I don't think there are other instances though.
### Impacted Areas

Anything that uses the regions macro/simple footer.
### Steps to Reproduce

The current footer on store front static has incorrect highlighting and spacing.  On this branch the highlighting and general styling more closely matches what we currently have for footer breadcrumbs.


josephco:
@ckeisato Looks good

PR #114, opened 2016-05-09
Switch from ava/nyc to mocha/chai

stewartyu:
### Issues

Moved to Mocha/Chai so that we can run PhantomJS tests.
### Todos
- [x] Get Nunjucks rendering to work
### Steps to Reproduce
1. Run `npm install` to update your node modules.
2. Run `npm test` to run tests.


src/accordion/Accordion.test.js
```diff
@@ -1,21 +1,75 @@
...
+};
+
+describe('Accordion', () => {
+  before(() => {
+    fixture.setBase('src');
+
+  });
+
+  beforeEach(() => {
+    accordion = fixture.load('accordion/Accordion.test.html');
```
ArmandoAmador: 👍

src/accordion/Accordion.test.js
```diff
@@ -1,21 +1,75 @@
...
+
+    assert.equal(Accordion.transitionDuration, 100);
+  })
+
+  it('should initialize with visible sectionContent', () => {
+    Accordion.init('#accordion');
+
+    const sectionContents = document.querySelectorAll('.js-accordion-section-content');
+
+    for (var i = 0; i < sectionContents.length; i++) {
```
ArmandoAmador: Just curious, would a `forEach` function work better here for the assertions of each sectionContent?

src/accordion/Accordion.test.js
```diff
@@ -1,21 +1,75 @@
...
+
+  });
+
+  beforeEach(() => {
+    accordion = fixture.load('accordion/Accordion.test.html');
+
+    document.body.insertAdjacentHTML('afterbegin', accordion);
+  });
+
+  afterEach(() => {
```
ArmandoAmador, line 18: What's the reason behind running this after each test?

ArmandoAmador:
Great catch on changing to Mocha/Chai this early and not after writing dozens of test @stewartyu 

Would we still consider Ava/NYC later in the future, or do you feel Mocha/Chai does everything we need?

src/accordion/Accordion.test.js
stewartyu: @ArmandoAmador Great question! NodeList's don't actually have a `forEach` function as it's not an Array: https://toddmotto.com/ditch-the-array-foreach-call-nodelist-hack/ I'm thinking for simplicity and familiarity, we can stick with the `for` loop.

src/accordion/Accordion.test.js
stewartyu, line 18: @ArmandoAmador The `beforeEach` is loading the accordion fixture HTML, so it's good to clean it up and reset it to the default in the `afterEach`, just in case there are any DOM changes.

stewartyu:
@ArmandoAmador If Ava adds support for browser testing with PhantomJS/Chrome, then we could definitely consider it again! 😄

stewartyu:
@misscs I've merged in the changes from `v2.0.0-dev` in here, but I'm running into an issue with the Nunjucks fixture rendering. It seems like some sort of Browserify issue that I haven't been able to figure out. In the meantime, I've reverted back to the flat HTML fixture for now, just so that I could confirm that the tests still work after the merge (and they do!). 

Do you want to pair on this when you have a chance?

josephco:
Looks good to me! 👍  DTM if others are cool with this.

PR #116, opened 2016-05-09
Add CasperUtilities and View JS files

ckeisato:
### Issues

Add Casper View.js and its dependency to nightshade-core
### Todos

Finalizing folder for shared JS files.
### Impacted Areas

FAQ page in storefront static
### Steps to Reproduce

None


misscs:
DTM. We may need to change how we import `casperutilities` so it is relative `./casperutilities` but we can make that change when we dig into it

PR #123, opened 2016-05-10
Move Phantom folder into src/

ckeisato:
### Issues

Move phantom into the src folder
### Todos

What's outstanding?
### Impacted Areas

### Steps to Reproduce


ckeisato:
verbal dtm from @josephco

PR #129, opened 2016-05-11
Fix reference to verlag-light

ckeisato:
### Issues

Rebase issue while merging v2.0.0 into master


ckeisato:
verbal dtm from @misscs

Issue #104, opened 2016-05-03
Create JS tests for Scrollto.js


misscs:
### Vision

Write tests for [Scrollto.js](https://github.com/CasperSleep/nightshade-core/blob/v2.0.0-dev/src/animation/ScrollTo.js) pattern using testing framework (#96). Achieve file code coverage > 90% (#102).
### Current State

No unit testing currently in `nightshade-core`. 
### Affected Areas

This pattern is used by
- [Beacon](https://github.com/CasperSleep/nightshade-core/tree/v2.0.0-dev/src/beacon)
- [Scrollto_nav](https://github.com/CasperSleep/nightshade-core/tree/v2.0.0-dev/src/scrollto_nav)

stewartyu:
Merged

PR #131, opened 2016-05-12
ScrollTo tests

stewartyu:
### Issues
#104 - ScrollTo.js tests

This is a basic functional test to ensure that it scrolls down on the page. 

There are a few issues with the `ScrollTo` plugin, but I'll open up separate issues for them. Tests can be expanded when more of these issues are sorted out.
### Steps to Reproduce

Run `npm test`.


karma.conf.js
```diff
@@ -89,7 +89,15 @@ module.exports = function(config) {
...
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['PhantomJS'],
+    browsers: ['Chrome_Sized'],
+
+
+    customLaunchers: {
+      'Chrome_Sized': {
+        base: 'Chrome',
+        flags: ['--window-size=400,600']
```
misscs, line 98: Does phantom have the ability to size the window?

src/animation/ScrollTo.js
```diff
@@ -11,6 +11,8 @@ export const ScrollTo = {
    * @param {number} [offset=0] - Scroll target offset
    * @param {number} [speed=1000] Speed of animation in ms
    * @returns {void}
+   *
+   * @todo Setting speed: 0 defaults back to 1000. Setting any speed value doesn't actually change visible speed.
```
misscs, line 15: @josephco Any thoughts on why the speed is not responding to passed in arguments?

karma.conf.js
stewartyu, line 98: @misscs It does - I just couldn't get it to work. Let me try again though because I do agree that would be the best way to go.

misscs, line 98: Sounds good. If you can't get it to work, just open up an issue to keep
track of what remains

Issue #130, opened 2016-05-12
Add Sass lint to pre-commit 


misscs:
### Vision

Want to run sass lint rules during pre-commit. 
### Current State

Linter is not running during pre-commit. 
### Scope
- [x] Add sass-lint.yml file
- [ ] Clean up linter errors

PR #132, opened 2016-05-12
Sass linting

stewartyu:
### Issues
#130 Added sass linting into `nightshade-core` and pre-commit hook.
### Todos
- [ ] Standardize linting rules
- [ ] Clean up 🐛 
### Steps to Reproduce
1. `npm install`
2. `npm run sasslint`

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,0 +1,103 @@
+#########################
+## Sample Sass Lint File
+#########################
+# Linter Options
+options:
+  # Set the formatter to 'html'
+  formatter: stylish
+# File Options
+files:
+  include: 'src/**/*.scss'
+# Rule Configuration
+rules:
+  # Extends
+  extends-before-mixins: 1
+  extends-before-declarations: 1
+  placeholder-in-extend: 1
+
+  # Mixins
+  mixins-before-declarations:
+    - 1
+    -
+      exclude:
+        - 'respond-at-and-above'
+        - 'respond-between'
+        - 'respond-below'
+        - 'after'
+        - 'before'
+
+  # Line Spacing
+  one-declaration-per-line: 1
+  empty-line-between-blocks: 1
+  single-line-per-selector: 1
+
+  # Disallows
+  no-color-keywords: 1
+  no-color-literals: 0
+  no-css-comments: 0
+  no-debug: 1
+  no-duplicate-properties: 1
+  no-empty-rulesets: 1
+  no-extends: 0
+  no-ids: 1
+  no-important: 1
+  no-invalid-hex: 1
+  no-mergeable-selectors:
+    - 1
+    -
+      whitelist:
+        - 'from'
+        - 'to'
+  no-misspelled-properties: 1
+  no-qualifying-elements: 1
+  no-trailing-zero: 1
+  no-transition-all: 1
+  no-url-protocols: 0
+  no-vendor-prefixes: 1
+  no-warn: 1
+  property-units: 0
+
+  # Nesting
+  force-attribute-nesting: 1
+  force-element-nesting: 1
+  force-pseudo-nesting: 1
+
+  # Name Formats
+  class-name-format: 1
+  function-name-format: 1
+  id-name-format: 0
+  mixin-name-format: 1
+  placeholder-name-format: 1
+  variable-name-format: 1
+
+  # Style Guide
+  bem-depth: 0
+  border-zero: 1
+  brace-style: 1
+  clean-import-paths: 1
+  empty-args: 1
+  hex-length: 1
+  hex-notation: 1
+  indentation: 1
+  leading-zero: 1
+  nesting-depth: 1
+  property-sort-order: 0
+  quotes: 1
+  shorthand-values: 1
+  url-quotes: 1
+  variable-for-property: 1
+  zero-unit: 1
+
+  # Inner Spacing
+  space-after-comma: 1
+  space-before-colon: 1
+  space-after-colon: 1
+  space-before-brace: 1
+  space-before-bang: 1
+  space-after-bang: 1
+  space-between-parens: 1
+  space-around-operator: 0
+
+  # Final Items
+  trailing-semicolon: 1
+  final-newline: 1
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -11,13 +11,13 @@ files:
 # Rule Configuration
 rules:
   # Extends
-  extends-before-mixins: 1
-  extends-before-declarations: 1
-  placeholder-in-extend: 1
+  extends-before-mixins: 2
+  extends-before-declarations: 2
+  placeholder-in-extend: 2
 
   # Mixins
   mixins-before-declarations:
-    - 1
+    - 2
     -
       exclude:
         - 'respond-at-and-above'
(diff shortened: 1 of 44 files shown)


stewartyu:
Atom users: I recommend installing the `linter-sass-lint` plugin which will highlight the linting errors in your editor!

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+        - 'before'
+
+  # Line Spacing
+  one-declaration-per-line: 2
+  empty-line-between-blocks: 2
+  single-line-per-selector: 2
+
+  # Disallows
+  no-color-keywords: 2
+  no-color-literals: 0
```
robwierzbowski: Difference between literals and keywords?

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  no-css-comments: 0
+  no-debug: 2
+  no-duplicate-properties: 2
+  no-empty-rulesets: 2
+  no-extends: 0
```
robwierzbowski: Would be nice to work towards this.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  no-ids: 2
+  no-important: 2
```
robwierzbowski, line 43: I find IDs and `!important` sometimes are necessary to manage third party code, but guessing there's an option for disabling linting per file / block / etc. In favor of keeping these errors, just pointing out.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  no-invalid-hex: 2
+  no-mergeable-selectors:
+    - 2
+    -
+      whitelist:
+        - 'from'
+        - 'to'
+  no-misspelled-properties: 2
+  no-qualifying-elements: 2
```
robwierzbowski, line 52: 🎉

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  no-trailing-zero: 2
+  no-transition-all: 2
```
robwierzbowski, line 54: Feedback from team desired on this one.

ckeisato:
Does this standardize the type of comments we use in sass files?

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  no-url-protocols: 0
+  no-vendor-prefixes: 2
```
robwierzbowski: What about vendor only rules like `-webkit-overflow-scrolling`? Autoprefixer won't prefix `overflow-scrolling`, but it's an essential rule to enable momentum scrolling on iOS.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  function-name-format: 2
+  id-name-format: 0
+  mixin-name-format: 2
+  placeholder-name-format: 2
+  variable-name-format: 2
+
+  # Style Guide
+  bem-depth: 0
+  border-zero: 2
+  brace-style: 2
```
robwierzbowski: Can we specifically set `stroustrup` and `allow-single-line: false`?

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  clean-import-paths: 2
+  empty-args: 2
+  hex-length: 2
+  hex-notation: 2
```
robwierzbowski: Does this default to lower or uppercase? I prefer enforcing lower.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  indentation: 2
+  leading-zero: 2
```
robwierzbowski: I'd like to switch this to require the leading `0`. I think it aids readability, and stays consistent with the other code we work with (JS, most of our Ruby). 

Sass / CSS minifiers will remove the leading `0` on minfication, so this is a purely developer experience decision.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+  nesting-depth: 2
+  property-sort-order: 0
+  quotes: 2
+  shorthand-values: 2
+  url-quotes: 2
+  variable-for-property: 2
```
robwierzbowski: Without an array of props to enforce, this rule doesn't do anything. Let's set it to 0 until we're actually using it.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+
+  # Inner Spacing
+  space-after-comma: 2
+  space-before-colon: 2
+  space-after-colon: 2
+  space-before-brace: 2
+  space-before-bang: 2
+  space-after-bang: 2
+  space-between-parens: 2
+  space-around-operator: 0
```
robwierzbowski: I'd like to enforce this rule.

.sass-lint.yml
```diff
@@ -0,0 +1,103 @@
...
+
+  # Line Spacing
+  one-declaration-per-line: 2
+  empty-line-between-blocks: 2
+  single-line-per-selector: 2
+
+  # Disallows
+  no-color-keywords: 2
+  no-color-literals: 0
+  no-css-comments: 0
```
robwierzbowski, line 37: I'd like to turn this on. Thoughts?

.sass-lint.yml
stewartyu: @robwierzbowski https://github.com/sasstools/sass-lint/blob/develop/docs/rules/no-color-literals.md

I'm not as familiar with our current Sass practices, but I do like the `no-color-literals` rule because it forces you to use our standard color functions/variables. For example, this is not allowed:

`background: 1px solid #asdf;`

Instead, you would need to do:

`$brand-color: #asdf;
background: 1px solid $brand-color;`

.sass-lint.yml
stewartyu, line 43: Agree - we should explicitly ignore `!important`s only when absolutely necessary, like when dealing with third-party CSS.

.sass-lint.yml
stewartyu, line 54: Due to performance drawbacks, I'd prefer to keep this enabled. You can explicitly enable multiple transitions like this:

`transition: color 0.2s, text-shadow: 0.2s;`

.sass-lint.yml
stewartyu: Interesting - I didn't know that Autoprefixer didn't prefix that one. How often does this happen - or is it only with this specific property? Could we potentially use a PostCSS plugin to handle this automatically for us? Or, we could individually ignore this rule.

.sass-lint.yml
stewartyu: How do we feel about `1tbs`? https://github.com/sasstools/sass-lint/blob/develop/docs/rules/brace-style.md#differences-between-1tbs-and-stroustrup

.sass-lint.yml
stewartyu: 👍 for lowercase.

.sass-lint.yml
stewartyu: 👍 keeping the leading 0.

.sass-lint.yml
stewartyu: 👍

.sass-lint.yml
stewartyu: 👍

.sass-lint.yml
robwierzbowski: Gotcha. I'm ambivalent on this rule (one off colors are a reality, and a one-off color variable is not necessarily useful), but don't have a strong feeling against it. I leave it up to others.

.sass-lint.yml
robwierzbowski, line 54: Are there perf implications if no transition is triggered? Asking for my own edification.

.sass-lint.yml
stewartyu: With the advent of `nightshade-core`, do we still envision one-off colours happening? Not sure of the exact direction of this, but you're right that one-off variables aren't useful.

.sass-lint.yml
robwierzbowski: I think it's overly complex to add a special prefixing module, and it would obfuscate the knowledge that it's a non-standard, vendor-only property. We do currently use that rule in Casper Rails, among others (`-moz` textfield ones, off the top of my head).

IMO, all of us are on the same page that autoprefixer handles prefixing, unless the property is non-standard. In those cases I want to see that we're using a non-standard property, so I would be in favor of taking this one out.

.sass-lint.yml
robwierzbowski: I hate it, haha.

I would like to move all JS to stroustrup too. Jam session topic.

robwierzbowski: For me, `if` and `else` have equal weight, it helps me scan, and the code symmetry + highlighting increases readability.

.sass-lint.yml
stewartyu, line 54: Not sure, actually. But I like the idea of forcing the developer to judiciously add transitions so that we don't get carried away with transitioning all when it's not actually necessary:

```
transition: color 1s, box-shadow 1s;
color: ...;
box-shadow: ...;
border: ...; // transition not required on border
font-weight: ...; // transition not required on font-weight
```

.sass-lint.yml
robwierzbowski: I think we can envision no one-offs, but there always are. I don't mind rolling them into a var, in the end. I see the benefit of naming to avoid the perception of "magic". 

@misscs?

.sass-lint.yml
robwierzbowski: If 5 props are `0.2s`, there comes a point where `all` increases readability, maintenance. Can you find out if the purpose of this linter is perf or style?

stewartyu:
@ckeisato Yes, if we enable `no-css-comments` as @robwierzbowski mentioned, then it would enforce CSS comment style. It wouldn't force the structure of the actual comment itself, but just the fact that we would always use `//` style comments instead of `/* */` style.

robwierzbowski:
@ckeisato Was your comment to me or Stewart?

.sass-lint.yml
stewartyu: Enable this flag.

.sass-lint.yml
stewartyu: Let's turn this on.

.sass-lint.yml
stewartyu, line 43: Agreed to keep these enabled.

.sass-lint.yml
stewartyu, line 54: We'll keep this one on.

.sass-lint.yml
stewartyu, line 37: Keep this off.

.sass-lint.yml
stewartyu: Let's keep this on and add `-webkit-overflow-scrolling` to `excluded-identifiers`.

.sass-lint.yml
stewartyu: Let's go with `stroustrup` and `allow-single-line: false`

.sass-lint.yml
stewartyu: Agreed on lowercase!

.sass-lint.yml
stewartyu: Keeping leading 0s!

.sass-lint.yml
stewartyu: Let's set this to 0.

.sass-lint.yml
stewartyu: Let's turn this on.

.sass-lint.yml
stewartyu: https://github.com/sasstools/sass-lint/blob/master/docs/rules/no-vendor-prefixes.md

PR #133, opened 2016-05-13
Move scale to a top level module, single file import

robwierzbowski:
### Issues

Currently modular scale is contained inside the typography directory. Modular scale itself is being used for more than type; padding, margin, spacing. In the rails app I'd like to import modular scale without typography mixins / changes, and be clear in the code that I'm doing so.

Currently we need to import two files in order to use modular scale. I'd like to be able to import a single file that contains the dependencies and logic for modular scale (e.g., `import 'scale/index'`). This matches our JS module pattern (and good modularity in general): self contained and easy to work with.
### TODO

After review, we should bump and release so we can pull into the Rails app.
### Impacted Areas

Modular scale is now loaded as a top-level module in Nightshade-core, and is available through a single import statement. So modular.
### Steps to Reproduce

Everything should work exactly the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@casper/nightshade-core",
-  "version": "2.0.1",
+  "version": "2.1.0",
   "description": "Casper pattern and style scss library. Under development.",
   "scripts": {
     "eslint": "eslint src",
--- a/src/_core_imports.scss
+++ b/src/_core_imports.scss
@@ -8,6 +8,7 @@
 @import './node_modules/susy/sass/susy';
 
 @import 'color/index';
+@import 'scale/config';
 
 @import 'layout/positioning_helpers';
 @import 'utilities/queries_helpers';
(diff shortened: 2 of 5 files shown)


robwierzbowski:
What do the `@file` comments mean at the top of the file? The originals didn't match the file name or the directory exactly (missing words, case), but I assume they have some relation?

robwierzbowski:
@josephco: Updated the PR with a version bump. Can you check it and give blessing / feedback?

robwierzbowski:
Added one additional commit for review. Seems like gulp sass reads from the location of the gulpfile first. I'm not positive how this will work with flattening of the dependency tree, but should we look at `./node_modules` for the import? Seems to work (although `../` also worked, which is a little disconcerting). 

What do people think is the safest way to write dependency paths as we pull this into multiple projects as a dependency? /cc @misscs.

josephco:
@robwierzbowski Looks good, version bump and all. I like the root path more, good move

robwierzbowski:
Revision commit squashed. Merging and tagging.

Issue #103, opened 2016-05-03
Create JS tests for Tooltip.js 


misscs:
### Vision

Write tests for Tooltip pattern using testing framework (#96). Achieve file code coverage > 90% (#102).
### Current State

No unit testing currently in `nightshade-core`.
### Scope
- [ ]

stewartyu:
PR: https://github.com/CasperSleep/nightshade-core/pull/135

stewartyu:
Merged

Issue #99, opened 2016-05-03
Create reusable set of JS helpers


stewartyu:
### Vision

Utility function helper for common methods.
### Current State

There is currently a lot of code duplication, implementing the same methods between different modules.
### Scope

These functions should be pulled out into a utility:
- [ ]

PR #135, opened 2016-05-13
Add tests for tooltip

stewartyu:
### Issues
#103 - tooltip tests
### Steps to Reproduce

Run `npm install` and `npm test`

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -89,7 +89,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['PhantomJS'],
+    browsers: ['Chrome'],
 
 
     // Continuous Integration mode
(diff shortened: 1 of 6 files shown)


karma.conf.js
```diff
@@ -89,7 +89,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['PhantomJS'],
```
misscs, line 92: Is this just for your testing purposes? In general, think PhantomJS would be better than Chrome since it is faster.

misscs:
✅ ✅ ✅ ✅ ✅ ✅ ✅ (to the tune of [cake](Rihanna - Birthday Cake %28Audio%29 - YouTube) 🍰 )

I'm curious to see what becomes of this test when you tackle Popovers. They are very similar and we had talked about abstracting out some of the common methods in #99. As is, DTM.

karma.conf.js
stewartyu, line 92: @misscs Yeah, I agree for sure. I switched this to Chrome because PhantomJS didn't seem to respect the viewport options that I set. Without the viewport settings, scrolling doesn't actually occur because it automatically makes the viewport large enough to fit all the page content.

PR #141, opened 2016-05-16
Remove karma-jasmine dev dep

stewartyu:
### Issues

We were never actually using Jasmine - this dependency slipped through by accident!
### Steps to Reproduce

Run `npm prune`, `npm install`, `npm test` and tests should still pass.


robwierzbowski:
DTM. No shrinkwrap in core, huh?

PR #142, opened 2016-05-18
Update fonts_config to have default fonts and include calibre

ckeisato:


src/phantom/base/typography/_bundle.scss
```diff
@@ -2,3 +2,15 @@
 @import "links_config";
 @import "lists_config";
 @import "paragraphs_config";
+
+@font-face {
+  font-family: "Calibre Light";
+  src: url('https://casper.com/static-assets/nightshade-core/src/typography/fonts/CalibreWeb-Light.woff2') format('woff2'),
```
misscs: If we want to go this route we may want to use the version from the Rails CDN which is the other AWS instance. 

Another option is to create a separate bucket just for fonts. This could work well in the case of self-hosting our typography.com files. 

@jpsilvashy can you weigh in here...we want to store our font files in a spot where we can access them across our apps. the best option for now is to reference them directly from the server. should we create a new bucket on the casper classic amazon aws account or the one you set up?

jpsilvashy: I think we're better off not including these in the asset pipeline on either service in that case, I think the best approach is a new bucket in the _new_ AWS account, where we link to those fonts from each app. Thoughts?

misscs: @jpsilvashy yes, i agree with you. i like the new bucket route and that's a clear path for us. when we reference the files in sass file, should we just use the aws url or cdn url?

misscs:
@josephco Here's PR that creates the fonts repo https://github.com/CasperSleep/nightshade-fonts/pull/1. Would be great to get CR from you, too.  

I'm still waiting to hear from JP about which URL you can use, but for now this path gives you https://s3.amazonaws.com/nightshade-fonts/fonts.min.css

PR #145, opened 2016-05-19
Allow header/footer take a different phone number

victorwhy:
Allows the simple header to take a different phone number, but keeps the current number as the default number


josephco:
@victorwhy Looks good

PR #148, opened 2016-05-23
Add SpriteFetcher.js to js_utilities

misscs:
Move this shared utility to central place for maintainability and shareability across
all projects

### DIFF
--- a/src/js_utilities/SpriteFetcher.js
+++ b/src/js_utilities/SpriteFetcher.js
@@ -1,0 +1,66 @@
+/**
+ * @overview Fetch an SVG spritesheet and append it to the DOM. This allows us
+ * to use `<use>` elements and target sprites with CSS in all modern browsers.
+ * @module SpriteFetcher.js
+ * @TODO: Use the new `fetch` API once Polyfill.io polyfills for ie9 and
+ * Safari correctly.
+ */
+
+export const SpriteFetcher = {
+  container: null,
+
+  /**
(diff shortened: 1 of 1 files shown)


victorwhy:
This doesn't need to be included in any calls / tasks here?  is this just so that it's included within this repo so that it's available natively?

misscs:
@victorwhy exactly. it will be called in the libraries. eventually i see us moving shared JS in `n-core` and may end up calling it. I will issue two PRs on `storefront-static` and `nightshade` with this update. Once I update our jspm useage in rails (as part of the js cleanup) we'll be able to use nightshade-core js in rails app.

misscs:
@victorwhy / @robwierzbowski good to go with this one? would be good to get this into the next release so i can update the other repos and open PRs.

robwierzbowski:
DTM. Are you going to make a PR to Casper Rails too?

robwierzbowski:
Wait! Can we hypenate the folder? I haven't seen us use underscored folders anywhere else in our apps.

robwierzbowski:
Huh, never mind, you do here. Commenting on a phone, takes a little while to research.

misscs:
We underscore for compatibility/sameness/convention with rails. Bough we
could make a case to switch later. I'm also ++ for hyphens

josephco:
DTM from me

robwierzbowski:
Consistency is all I care about! DTM!

PR #149, opened 2016-05-24
Remove H1 top margin added by normalize.scss

josephco:
This is a patch that will update to v2.5.1, fixing an issue where the latest version of normalize.css adds top margins to H1 elements.


robwierzbowski:
DTM. 

IMO we should do a hard reset on our type and list styles to avoid this in the future, /cc @misscs, @stewartyu.

stewartyu:
@robwierzbowski Agreed! If there's any normalization that we like from `normalize`, I think it's more maintainable to apply those on top of a reset.

PR #150, opened 2016-05-24
Hotfix feature.js import in Toggler

misscs:


misscs:
Closing. Need to fix this at the JSPM level in Nightshade

PR #151, opened 2016-05-26
Un-interpolate return value of size function

stewartyu:
### Issues

The `size` function returns a string, which makes it difficult to do math in Sass. I'm un-interpolating the return value so that you can do things like this now:

`size(largest) * -1`
### Steps to Reproduce
1. Run `npm link` in `nightshade-core`.
2. Run `npm link @casper/nightshade-core` in Casper Rails.
3. Make sure that the above syntax works, and that the `size` function is returning correctly.

### DIFF
--- a/src/phantom/base/_sizing_config.scss
+++ b/src/phantom/base/_sizing_config.scss
@@ -60,5 +60,5 @@ $sizing: (
 
   $size: map-get($sizing, $size) or $size;
 
-  @return #{$size};
+  @return $size;
 }


victorwhy:
i'm a little curious on why we interpolated in the first place.  what is `0.25em` if not a string?

stewartyu:
@victorwhy Even though `0.25em` has a unit, it's considered to be a `Number`: http://sass-lang.com/documentation/Sass/Script/Value/Number.html

robwierzbowski:
@victorwhy Sass has a concept of numbers with units. So it can add `1px + 10px` but knows not to add `1px + 5em`.

robwierzbowski:
A+! DTM.

PR #152, opened 2016-05-26
Fix issue where Chronicle Display shows instead of Deck

josephco:


victorwhy:
why only for that breakpoint? is it literally a differnet chronicle for bigger screens?

josephco:
@stewartyu We only use Chronicle Display on small screens and switch in Chronicle Deck on larger ones

PR #154, opened 2016-05-31
Use arrow-down in accordion

ckeisato:
### Issues

The arrow icon was being used instead of the arrow-down icon.  This PR replaces the arrow with the correct one and adjusts the styling accordingly.
### Todos

None.
### Impacted Areas

The accordion module in nightshade and in the Casper Rails repo.
### Steps to Reproduce

None.


robwierzbowski:
DTM.

PR #155, opened 2016-05-31
Use arrow-down icon

ckeisato:
I didn't update the icon in my last pr, also wasn't able to catch this because I was having issues running nightshade yesterday.  Sorry.


robwierzbowski:
DTM

PR #158, opened 2016-06-13
Footer

misscs:
- Prepare footer for international
- Improve/simplify markup

### DIFF
--- a/src/footer/footer.macros.html
+++ b/src/footer/footer.macros.html
@@ -7,53 +7,42 @@
 {#
   * Renders a standard or simplified footer
 #}
-{% macro footer(simple=false, class=null, id=null, phone_number='888–498–0003') %}
+{% macro footer(data, simple=false, class=null, id=null) %}
 <footer class="footer{% if simple %} footer--simple{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
   <div class="footer-top r-inner">
     {{ caller() }}
   </div>
   <div class="footer-bottom">
-    <nav class="footer-nav r-inner">
-      <div class="footer-nav-group footer-nav-group--bottom">
-        <span class="footer-nav-country-icons">
-          <a class="footer-nav-group-link footer-nav-group-link--country" href="?country_id=49">
-            {{ icon('usa', size='small') }}
+    <div class="footer-bottom-inner r-inner">
+
+      <span class="footer-bottom-item site-copyright">
+        &copy;
+        <script>document.write(new Date().getFullYear());</script>
+      </span>
+
+      <span class="site-legal-links footer-bottom-item">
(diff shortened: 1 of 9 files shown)


src/footer/footer.macros.html
```diff
@@ -7,53 +7,42 @@
 {#
   * Renders a standard or simplified footer
 #}
-{% macro footer(simple=false, class=null, id=null, phone_number='888–498–0003') %}
+{% macro footer(data, simple=false, class=null, id=null) %}
 <footer class="footer{% if simple %} footer--simple{% endif %}{% if class %} {{ class }}{% endif %}"{% if id %} id="{{ id }}"{% endif %}>
   <div class="footer-top r-inner">
     {{ caller() }}
   </div>
   <div class="footer-bottom">
```
misscs, line 15: What's a better class to replace `footer-bottom`? I like the feel of the the `site-` classes to describe the content, but it now feels disconnected from its parent class.

src/footer/footer.macros.html
```diff
@@ -7,53 +7,42 @@
...
+        <a class="" href="https://casper.com/privacy">{{ data.privacy }}</a>
+        <a class="" href="https://casper.com/terms">{{ data.terms }}</a>
+      </span>
+
+      <span class="site-address h-card footer-bottom-item">
+        <span class="p-street-address">{{ data.address }}</span>
+
+        {% if not simple %}
+        <span class="site-address-meta">
+          <a class="footer-bottom-item" href="tel:{{ data.phone_number|replace('-', '') }}">
```
misscs, line 33: Not so hot on `footer-bottom-item` either, but I need a class to target each `item` for spacing. This may change once design looks at this...less urgent now.

misscs:
@CasperSleep/frontier Left a couple comments looking for some input on class names. Thank you!

src/footer/footer.macros.html
stewartyu, line 15: I'm not a huge fan of class names that describe position, since the positions could theoretically be made incorrect in certain media queries via CSS. `footer-bottom` seems like navigation - could this be `footer-nav`?

src/footer/footer.macros.html
stewartyu, line 33: If we change `footer-bottom` to `footer-nav`, then `footer-nav-item` would actually work pretty well.

robwierzbowski, line 33: Agree, no positional names in classes please.

robwierzbowski, line 33: Footer bottom seems a little redundant as well `:)`

misscs, line 33: Agreed re positional names. I'd like to find a name that reflects the
content. `footer-nav` feels like a better class to replace `footer-top`.
What is now the bottom is not really a nav. It is site/company
details—copyright, address, contact, and links to terms and privacy.
Looking for a name to encompass the area that hold site/company
information.

PR #159, opened 2016-06-13
Add extends for calibre light

ckeisato:
Remove all font extends

### DIFF
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -241,7 +241,7 @@ select {
 }
 
 .form-label--alt {
-  @extend %verlag-black;
+  @include verlag-black;
 
   font-size: 15px;
   text-transform: uppercase;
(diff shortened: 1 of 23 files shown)


josephco:
This looks like a fine solution to me. Let's see what the others think.

josephco:
This might be a good opportunity to convert our font extends to mixins, the code is starting to feel a little unwieldy. If we need to re-declare Calibre Light for all of these font definitions, let's leave a @todo to remove them later once we are fully integrated.

src/forms/_base.scss
```diff
@@ -228,9 +228,8 @@ select {
 
 
 .form-label {
-  @extend %headers;
-
-  @extend %h4;
+  @include headers;
```
josephco: Can probably delete this line, but test it.

josephco:
LGTM, DTM but maybe get @robwierzbowski or @stewartyu to give it a once-over.

robwierzbowski:
Nice. The compiler should catch any missing extend calls, and the diff looks great. DTM.

stewartyu:
Sorry, I was looking at this earlier today but got sidetracked. LGTM!

PR #161, opened 2016-06-17
Add margins to heading classes, make primary blue

josephco:
This should be a patch version and updated on Casper repo

### DIFF
--- a/src/typography/_headings_config.scss
+++ b/src/typography/_headings_config.scss
@@ -1,48 +1,40 @@
-@mixin headers {
+@mixin heading {
   @include verlag-black;
   text-transform: uppercase;
   line-height: 1.4;
   color: color(primary);
 }
 
-@mixin headers--dark {
-  color: color(primary);
-}
-
-@mixin headers--light {
-  color: color(background);
-}
-
 @mixin heading-1 {
-  @include headers;
+  @include heading;
   font-size: ms(4);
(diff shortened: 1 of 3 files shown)


misscs:
Nice to see this come together and get some sensible defaults.

src/typography/_headings_config.scss
```diff
@@ -2,6 +2,7 @@
   @include verlag-black;
   text-transform: uppercase;
   line-height: 1.4;
+  color: color(primary);
 }
 
 @mixin headers--dark {
```
misscs: do we need this? doesn't seem like a concept we need to carry over from old architecture

src/typography/_headings_config.scss
```diff
@@ -2,6 +2,7 @@
   @include verlag-black;
```
misscs, line 2: since we are using `heading` as the base class. can we change this mixin to follow the same terminology `heading` or `headings`

misscs:
@josephco DTM pending couple comments

src/typography/_headings_config.scss
josephco, line 2: @misscs I'll call this `@mixin heading`

src/typography/_headings_config.scss
josephco: Pulling out these color modifiers

PR #164, opened 2016-06-18
PDP Fit & Finish

josephco:


src/layout/_page.scss
```diff
@@ -14,3 +14,7 @@
     padding: ms(10) ms(2);
   }
 }
+
+.page-section--compact {
```
stewartyu: Should we be chaining modifiers onto their module? i.e. `.page-section.page-section--compact { ... }` to increase specificity.

robwierzbowski: Agree, chain or nest. We don't do this everywhere now. @misscs, Can we get an opinion?

robwierzbowski:
Overall, LGTM. I would be fine merging as is, no matter the result of that style discussion. Let's give it a few mins and see if Claudina adds an opionon.

src/layout/_page.scss
josephco: @robwierzbowski @stewartyu 

In this case the modifier is used on its own even though it’s a “modifier", you would never use the two classes together. If not using a modifier, I couldn't think of what the class name should be... maybe just `page-section-compact`. It is a variant of the page-section class, which should be the default for most modules.

misscs: @robwierzbowski @stewartyu @josephco My opinions are the following:
1. If we want to increase specificity, I like nesting with `&` because I find it more readable. 
2. We don't currently chain our BEM modifiers.
3. @josephco I think your change to element from modifier is much clearer

robwierzbowski: 👍  Let's Jam on it later then.

robwierzbowski:
DTM.

src/layout/_page.scss
josephco: I renamed from modifier to element.

PR #167, opened 2016-06-24
Update page-section classes, Trial/Returns SCSS

josephco:
Make compact page section the default, add large
Move Trial/Returns IE styles from Casper

### DIFF
--- a/src/trial_returns/trial.macros.html
+++ b/src/trial_returns/trial.macros.html
@@ -22,10 +22,10 @@
     </div>
   {% endif %}
 
-  <div class="trial-returns-content page-section r-inner">
+  <div class="trial-returns-content page-section-large r-inner">
     {{ icon.svg(data.icon) }}
-    <h1 class="heading-1 trial-returns-heading">{{ data.heading }}</h1>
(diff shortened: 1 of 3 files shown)


src/trial_returns/trial.macros.html
```diff
@@ -22,10 +22,10 @@
     </div>
   {% endif %}
 
-  <div class="trial-returns-content page-section r-inner">
+  <div class="trial-returns-content page-section-large r-inner">
     {{ icon.svg(data.icon) }}
-    <h1 class="heading-1 trial-returns-heading">{{ data.heading }}</h1>
-    <p class="trial-returns-copy body--serif">{{ data.copy }}</p>
+    <h1 class="trial-returns-heading heading-1">{{ data.heading }}</h1>
```
ArmandoAmador, line 27: Why a number? for `heading-1`

josephco, line 27: @ArmandoAmador We made new heading classes to move the styling away from targeting the h1-h6 elements, and instead allow us to use a large (heading-1) style on any H1-6 element (this is better for keeping SEO/style separate)

PR #171, opened 2016-06-29
Fix Trial/Returns background image bug on Firefox

josephco:


robwierzbowski:
DTM!

robwierzbowski:
@josephco What was the issue, out of curiosity? Funny we have a foreground image called "background-image".

josephco:
@robwierzbowski It's a div with a picture element that behaves as a background image (content sits above it). In firefox the image is getting cut off, so it needed width: 100%

PR #174, opened 2016-07-13
Update Play button to be flexible and use rems

josephco:


src/buttons/_base.scss
```diff
@@ -159,38 +159,34 @@
 
 [data-button="play"] {
   position: relative;
-  display: inline-block;
-  min-height: 56px;
-  border-radius: 35px;
+  min-height: 3.5rem;
+  border-radius: 2.188rem;
```
misscs: Need to get this documented and in the linter rules, but what is our precision standard?

src/buttons/_base.scss
```diff
@@ -159,38 +159,34 @@
...
   @include after() {
-    @include arrow(right, color(primary), 10px);
-    left: 24px;
-    top: 31px;
-    border-top-width: 6px;
-    border-bottom-width: 6px;
+    @include arrow(right, color(primary), 0.625rem);
+    left: 1.5rem;
+    margin-top: -0.375rem;
+    border-top-width: 0.375rem;
```
misscs: Nice. Are we moving border widths over to ems?

src/buttons/_base.scss
josephco: @misscs I think we talked about 2 or 3 places. I like to do 3 for precision, but not sure if it makes much difference.

src/buttons/_base.scss
josephco: @misscs I'm not sure, just changed everything to rems, but when we do 1px borders we seem to leave them in pixels. I can keep them in pixels for consistency if that's preferred.

src/buttons/_base.scss
misscs: Would be great to firm up the answer to find out if we need to round this :) I recommend asking the team.

src/buttons/_base.scss
misscs: At the moment all the borders are pixels so +1 for consistency. Feel free to bring up change at the jam though! We want to get input before implementing a change from convention.

src/buttons/_base.scss
josephco: @misscs People in squad_who_code seem to be happy with 2 places

src/buttons/_base.scss
josephco: @misscs Sure, sounds good to me!

PR #176, opened 2016-07-25
Update Forms base SCSS with latest styles

josephco:
Updates Forms styles to be included in core_config_nightshade in Casper repo

### DIFF
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -1,51 +1,8 @@
-//doc
-//
-// ---
-// title: Forms
-// name: forms
-// category: components
-// ---
-//
-// Casper form components.
-//
-// Standard form components are always preferred over custom elements.
-//
-//
-//  Class            |  Description
-//  ---------------- | -----------------
-//  `.form-field`    | The element that wraps label and input
-//  `.form-legend`   | Title of the form
-//  `.form-label`    | The element class for form label.
-//  `.form-label-copy`     | Class to use on `span` element when you need to wrap text
-//  `.form-input`    | The element class for form input
-//  `.form-help`     | Class to designate form or form field help text
-//  `.form-actions`  | The element that contains form buttons and actions
-//
-//
-// ```haml_example
-// %form
-//
-//  %fieldset
-//
-//    .form-legend I am a form
-//
-//    .form-field
-//      %label.form-label
-//        Size
-//      %input{:type => 'text', :class => 'form-input'}
-//
-//    .form-field
-//      %span.form-label Price
-//      %input{:type => 'text', :class => 'form-input'}
-//
-//      %p.form-help Financing is available
-//
-//    .form-actions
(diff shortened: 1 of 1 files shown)


src/forms/_base.scss
```diff
@@ -173,30 +120,26 @@ select {
...
+  @include respond-at-and-above($form-max) {
     display: flex;
   }
 }
 
 .form-legend-alt {
-  @include body--sans-small;
-
+  font-size: ms(0);
+  letter-spacing: size(smaller-tracking);
```
ckeisato, line 137: why do you need to state the letter spacing here?  it doesn't inherit an appropriate value from the body?

src/forms/_base.scss
```diff
@@ -219,18 +158,77 @@ select {
...
+    top: ms(-4);
+    left: 1.1rem;
+    font-size: ms(-1);
+    letter-spacing: size(smaller-tracking);
+    line-height: 1;
+    color: color(dark-text);
+    opacity: 0;
+    pointer-events: none;
+    transform: translateY(ms(-9));
+    transition: transform 0.25s ease-out, opacity 0.25s ease-out;
```
ckeisato, line 226: didn't we have a transition mixin?  or a transition variable?

src/forms/_base.scss
```diff
@@ -345,17 +365,17 @@ select {
...
+  font-size: ms(-1);
+  color: color(primary);
+  margin-top: 0.7rem;
 
   a {
-    @include link(text, medium-text, "after");
+    @include link(text, dark-text, "after");
+  }
+
+  &.form-help--tight {
```
ckeisato, line 377: what is the `--tight` version for?

src/forms/_base.scss
josephco, line 226: Not sure, we usually just write it out like this.

src/forms/_base.scss
josephco, line 377: Not sure, this was carried over from casper repo, prob where there needed to be less margin in some instance.

src/forms/_base.scss
```diff
@@ -173,30 +120,26 @@ select {
...
     display: flex;
   }
 }
 
 .form-legend-alt {
-  @include body--sans-small;
-
+  font-size: ms(0);
+  letter-spacing: size(smaller-tracking);
   text-transform: lowercase;
```
ckeisato, line 138: why is this necessary?

src/forms/_base.scss
josephco, line 137: This is overwriting styles because it's nested within an H2

ckeisato:
looks good, is there a corresponding forms ie stylesheet that should be updated?  just noticed you added some flexbox stuff.

src/forms/_base.scss
josephco, line 138: Overwrites H2 styles

PR #179, opened 2016-09-06
Add styling changes for required fields change.

ryanhagerty:
Related to PR https://github.com/CasperSleep/Casper/pull/4790.

This changes instances where the required label was hidden until a user starts typing.

This removes the placeholder text and displays the label upfront, animating its position via transform once a user starts typing.

I'm sure you'll notice the use of `-webkit-autofill`. The reasoning for this is a nice-to-have that adds a little polish to browser autocomplete. Functionally, it works fine without these styles, but in the instance where a user starts the autocomplete process and hovers over a saved autocomplete address (before clicking), the styling looks like:

screen shot 2016-09-06 at 11 54 22 am

Because this is only visible when a user hovers on autocomplete and before clicking, I think it's fine for other browsers. Other sites that follow similar design patterns with labels, notably Warby Parker, still possess a similar issue (autocompleted info is cut off by background): 

screen shot 2016-09-06 at 11 49 10 am

Vendor prefixes aren't ideal, but it does make it look and read a lot better, even if it's a very short in-between state of selecting an autocomplete address:

screen shot 2016-09-06 at 11 53 42 am

### DIFF
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -202,36 +202,34 @@ select {
   &.has-text {
 
     .form-label {
-      transform: translateY(0);
-      opacity: 1;
+      transform: translateY(0.45rem) scale(0.8);
       pointer-events: auto;
     }
 
     .form-input {
       padding: 1.3rem ms(0) 0.4rem;
     }
-
-    .form-input--select {
-      color: color(text);
-    }
   }
 
   .form-label {
     position: absolute;
-    top: ms(-4);
-    left: 1.1rem;
-    font-size: ms(-1);
+    top: 0;
+    font-size: ms(0);
     letter-spacing: size(smaller-tracking);
     line-height: 1;
     color: color(dark-text);
-    opacity: 0;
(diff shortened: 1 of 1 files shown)


josephco:
When going to the forms I don't see any label text, it's invisible until text is input. It should be visible by default, and then shrink up above the input text on input.
screen shot 2016-09-06 at 5 07 37 pm

src/forms/_base.scss
```diff
@@ -202,36 +202,29 @@ select {
...
-    transition: transform 0.25s ease-out, opacity 0.25s ease-out;
+    transform: translateY(1rem) scale(1);
+    transition: transform 0.25s ease-out;
+    transform-origin: 0 0;
   }
 
-  .form-input--select {
-    color: color(dark-text);
+  .form-input {
+    transition: padding 0.15s ease-out;
```
josephco: We can remove this, now that the `input` event speeds things up

PR #180, opened 2016-09-08
Styling for iOS checkbox/radio

stewartyu:
### Issues

Fixing checkboxes/radios to use default iOS styling

Trello: https://trello.com/c/bOtRmlwl/239-1-checkout-remove-checkmark-icon-for-mobile
### Steps to Reproduce
1. `npm link` in `nightshade-core`
2. `npm link @casper/nightshade-core` in Casper Rails
3. Go to billing page on iOS
### Updated styling

screen shot 2016-09-08 at 1 47 22 pm
screen shot 2016-09-08 at 1 47 27 pm
### Old styling

screen shot 2016-09-08 at 1 48 11 pm
screen shot 2016-09-08 at 1 48 15 pm

### DIFF
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -57,6 +57,10 @@ input[type="radio"] {
   -webkit-appearance: radio;
   border-radius: 50%;
   border: 1px solid color(text);
+
+  &:checked {
+    background-color: color(gray);
+  }
 }
 
 // Get rid of number dropdown "spinners" for browsers
@@ -334,11 +338,18 @@ select {
 
 .form-input--checkbox {
   width: auto;
-  border: 0;
-  border-radius: 0;
   margin: 0;
   padding: 0;
   cursor: pointer;
+
+  // iOS styling
+  background: color(light-background);
+  border: 1px solid color(gray);
+  border-radius: 5px;
+
+  &:checked {
+    background: color(gray);
+  }
 }
 
 .form-input--radio {


src/forms/_base.scss
```diff
@@ -57,6 +57,10 @@ input[type="radio"] {
   -webkit-appearance: radio;
   border-radius: 50%;
   border: 1px solid color(text);
+
+  &:checked {
+    background-color: color(gray);
```
victorwhy, line 62: is `gray` the right color to use? i think this just uses the "regular" browser `gray`.  unless that's what we want

src/forms/_base.scss
```diff
@@ -334,11 +338,18 @@ select {
...
 .form-input--checkbox {
   width: auto;
-  border: 0;
-  border-radius: 0;
   margin: 0;
   padding: 0;
   cursor: pointer;
+
+  // iOS styling
+  background: #eee;
```
victorwhy: what color is `#eee`? also within our color map?

src/forms/_base.scss
stewartyu: `gray` actually becomes `#505050` https://github.com/CasperSleep/nightshade-core/blob/master/src/color/lib/_config.scss#L7

src/forms/_base.scss
stewartyu: Oh good catch, I meant to change that to use our color map.

PR #181, opened 2016-10-06
New checkbox styling

stewartyu:
### Issues

New checkbox styling:

screen shot 2016-10-06 at 10 19 56 am
screen shot 2016-10-06 at 10 20 03 am
screen shot 2016-10-06 at 10 20 09 am
### Steps to Reproduce
1. Go to http://localhost:3000/de/de
2. Add an item to cart
3. Go to shipping page to see default styling
4. Select the checkbox to see selected styling
5. Submit form with terms of service opt-in de-selected to see error styling

### DIFF
--- a/src/forms/_checkbox.scss
+++ b/src/forms/_checkbox.scss
@@ -1,0 +1,36 @@
+.checkbox {
+  display: none;
+
+  + .checkbox__label {
+    @include before() {
+      @include square(1rem);
+      top: 4px;
+      left: 0;
+      border: 1px solid color(gray--light3);
+      cursor: pointer;
+      text-align: center;
+      line-height: 0.5;
+    }
+  }
+
+  &.error {
+    & + .checkbox__label {
+      @include before() {
+        border-color: color(accent);
+      }
+    }
+  }
+
+  &:checked {
+    border-color: color(blue);
+
+    & + .checkbox__label {
+      @include before() {
+        content: '\d7';
+        color: color(primary);
+        border-color: color(primary);
+        font-size: ms(2);
+      }
+    }
+  }
+}
--- a/src/forms/_index.scss
+++ b/src/forms/_index.scss
@@ -5,3 +5,4 @@
 @import "toggle_locale";
 @import "accounts";
 @import "input_group";
+@import "checkbox";


src/forms/_checkbox.scss
```diff
@@ -0,0 +1,33 @@
+.checkbox {
+  display: none;
+
+  + .checkbox__label::before {
```
josephco: We should decide if we prefer single or double colon notation for pseudo elements. Probably better to maintain consistency for now and just use one?

src/forms/_checkbox.scss
```diff
@@ -0,0 +1,33 @@
...
+    content: '';
+    position: absolute;
```
josephco: We have a `@include before()` mixin that will cover the `content: ''` and `position: absolute` values:
https://github.com/CasperSleep/nightshade-core/blob/master/src/phantom/utilities/_visual_utilities.scss#L9-L18

Example of usage: https://github.com/CasperSleep/Casper/blob/master/app/assets/stylesheets/modules/products/_specs.scss#L18-L22

src/forms/_checkbox.scss
```diff
@@ -0,0 +1,33 @@
...
+    top: 4px;
+    left: 0;
+    width: 16px;
```
josephco: For equal width/height, we can use our `@include square(1rem)` mixin.

src/forms/_checkbox.scss
```diff
@@ -0,0 +1,33 @@
...
+      border-color: color(red);
+    }
+  }
+
+  &:checked {
+    border-color: color(blue);
+
+    & + .checkbox__label::before {
+      content: '\d7';
+      color: color(blue);
```
josephco: Our blue should be specified as `primary` to avoid confusion with the default CSS color named values.

src/forms/_checkbox.scss
```diff
@@ -0,0 +1,33 @@
...
+      border-color: color(blue);
+      font-size: 22px;
```
josephco: Can we use an `ms()` value such as `ms(2)` (23px)? If not, we should specify the font in rems (1.375rem in this case).

stewartyu:
Thanks! Addressed your comments - let's use single colon for consistency for now.

josephco (approved):

PR #183, opened 2016-10-24
Form label less emphasis

stewartyu:
### Issues

Adding a `.form-label--less-emphasis` modifier.

screen shot 2016-10-24 at 1 50 30 pm

### DIFF
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -195,6 +195,19 @@ select {
   & + .form-input {
     margin-top: ms(-6);
   }
+
+  &.form-label--less-emphasis {
+    @include calibre-medium;
+    font-size: ms(0);
+    letter-spacing: size(smaller-tracking);
+    text-transform: none;
+
+    &.form-label--radio-check {
+      .form-input--checkbox {
+        top: 6px;
+      }
+    }
+  }
 }
 
 .form-field--float {
--- a/src/forms/_base.scss
+++ b/src/forms/_base.scss
@@ -184,20 +184,16 @@ select {
 .form-field {
   position: relative;
   margin-bottom: ms(3);
+}
 
-  // @todo Turn this into a modifier when nightshade styles port over (e.g. form-field--standard)
-  &:not(.form-field--float) {
+.form-label {
+  @include heading-5;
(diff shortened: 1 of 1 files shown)


src/forms/_base.scss
```diff
@@ -184,18 +184,27 @@ select {
...
-      display: block;
-      margin-bottom: ms(-6);
-      color: color(text);
+  & + .form-input {
+    margin-top: ms(-6);
+  }
 
-      & + .form-input {
-        margin-top: ms(-6);
+  &.form-label--mixed-case {
```
robwierzbowski: It seems like the purpose of this is to be lower in hierarchy than a title case label. Can the modifier name reflect that?

src/forms/_base.scss
```diff
@@ -226,6 +235,11 @@ select {
     transform: translateY(1rem) scale(1);
     transition: transform 0.25s ease-out;
     transform-origin: 0 0;
+    text-transform: none;
+
+    + .form-input {
```
robwierzbowski: Can you use the `&` here, as on line 195?

stewartyu: I'd actually prefer to remove the `&` on line 195 as it's not actually necessary. I prefer only using the `&` when you need to change the default nesting order, as in `.foo &`. Thoughts?

src/forms/_base.scss
stewartyu: That's fair - any thoughts? Some of my original thoughts were `less-emphasis`, `level-2`, `lower` ... do we like any of these better?

src/forms/_base.scss
robwierzbowski: I think it's standard in the Casper codebase to use the `&` explicitly except when doing simple descendent nesting. I would be down to talk about a global refactor.

src/forms/_base.scss
robwierzbowski: `subheading`... I like all of these better than mixed case. I leave it up to you.

stewartyu:
@robwierzbowski Comments addressed!

robwierzbowski:
DTM!

PR #188, opened 2017-01-24
V2.19.0

misscs:
### Issues

Assortment of improvements for nightshade core that affects the storefront-static repo

* Remove utility classes on macros to make them more flexible. 
* Update footer to work for all countries. 
* Some IE fixes

I have additional PRs for strorefront static to bring it up to date with production.

### Todos

- [x] Write up testing steps for storefront static
- [x] Code/Review Testing
- [ ] Bump version

### Impacted Areas

- All headings except Upgrade section heading 
- Value props
- Footer

### Known Issues

Until updates are made to storefront static the following modules are not in parity with the live site. Most have issues at certain breakpoints. 

**All browsers**

- Upgrade callout missing heading
- Footer links and text are missing

**IE only**

- Awards
- Value props at table-ish sizes

#### On Storefront Static

These are screenshots of `master` branch without and with the changes in this branch changes. Areas that appear to be regressions from current live site will be addressed in separate commits to `storefront-static`. 

##### Chrome 

**Before/without patch (current state of master)**
ease-before-patch

**After/with patch**
ease-after-patch

##### IE

**IE Before**
ease-ie-before-patch

**IE After**
ease-ie-after-patch

To test this patch on Storefront static, create a link as outlined in directions below 👇 

#### Ensure no regressions on Casper.com Rails repo

- Change to this directory and branch. 
  - Run `nvm current`. 
  - If version is not `5.10.1`, `nvm use 5.10.1`.
- Create a link `npm link`
- Link repo to Casper Rails site `npm link @casper/nightshade-core`
- Click through PDPs, there should be any regressions

PR #189, opened 2017-01-25
V2.20.0 webpack et al

misscs:
### Issues

Updates to patterns to support storefront static v3 update. 

This PR is based on #188. You could easily choose to combine them in one release. Last 3 commits are the new ones. 

### Todos

- [ ] Testing

### Impacted Areas

- Footer
- CTA Banner

### Steps to Reproduce

### DIFF
--- a/src/layout/regions.macros.html
+++ b/src/layout/regions.macros.html
@@ -181,16 +181,19 @@
         {{ icon.svg("logo")}}
         <span class="nav-item-text is-invisible">Home</span>
       </a>
-      <ul class="nav-main-group nav-main-group--user">
-        <li role="presentation">
-          <span class="phone-number js-a-sup-phone">
-            <a class="nav-item" data-track-click-target="phone" href="tel:1{{ phone_number|replace('-', '') }}">
-              {{ icon.icon("phone")}}
-              {{ phone_number }}
-            </a>
-          </span>
-        </li>
-      </ul>
+
+      {% if phone_number %}
+        <ul class="nav-main-group nav-main-group--user">
+          <li role="presentation">
+            <span class="phone-number js-a-sup-phone">
+              <a class="nav-item" data-track-click-target="phone" href="tel:1{{ phone_number|replace('-', '') }}">
+                {{ icon.icon("phone")}}
+                {{ phone_number }}
+              </a>
+            </span>
+          </li>
+        </ul>
+      {% endif %}
     </nav>
   </div>
 </header>
--- a/src/value_props/_base.scss
+++ b/src/value_props/_base.scss
@@ -4,6 +4,11 @@
  * @overview Value props with side by side content and images
  */
 
+.value-prop {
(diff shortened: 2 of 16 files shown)


misscs:
@robwierzbowski Another lil PR. It could easily become part of #188. I left it separate to avoid side effects since this repo has changes to support the move from jspm to webpack in storefront static (see https://github.com/CasperSleep/nightshade-core/pull/189/commits/eed60382ff97212a72734f54a6ddc12c5cd47c31)

src/layout/regions.macros.html
```diff
@@ -181,16 +181,19 @@ <h4 class="footer-nav-group-title -light">Read More</h4>
...
-              {{ icon.icon("phone")}}
-              {{ phone_number }}
-            </a>
-          </span>
-        </li>
-      </ul>
+
+      {% if phone_number %}
+        <ul class="nav-main-group nav-main-group--user">
+          <li role="presentation">
```
robwierzbowski, line 187: Why role presentation here?

src/value_props/_base.scss
```diff
@@ -55,6 +60,28 @@
 
 .value-prop-heading {
   @include margin-auto;
+  @include heading-4;
   margin-bottom: ms(-1);
-  font-size: ms(0);
+
```
robwierzbowski: extra line break

src/value_props/_base.scss
```diff
@@ -55,6 +60,28 @@
...
+}
+
+// Variant: Prop without an image
```
robwierzbowski: Same comment as on the other PR :). Do these have overlap?

src/value_props/_ie_styles.scss
```diff
@@ -16,3 +16,9 @@
     margin: 0 7vw 0 0;
   }
 }
+
+
```
robwierzbowski: DLB

robwierzbowski (commented):
Seems pretty clear! One question and a couple extra line breaks.

src/layout/regions.macros.html
misscs: To be honest, I don't think this needs a complete ul structure just for the phone number. We used role presentation here because in this context list isn't really used for its semantic meaning. 

Shave some bytes and just get rid of the ul structure all together

src/value_props/_base.scss
misscs: Yea, this is based on that. I'll rebase it in.

misscs:
@robwierzbowski Question answered and whitespace removed. Let me know if you see the value in the code. Otherwise, I'll keep an eye out for your version number bump and update storefront-static

src/layout/regions.macros.html
stewartyu: [Rob] Fine for now!

stewartyu:
[Rob] Mergin!

PR #193, opened 2017-03-09
Remove reference to missing icon

spanuska:
### Issues

Removes reference to non-existent icon

### Steps to Reproduce

The way I found it was in the Casper repo, if you go to checkout and get to the address form, you'll notice a 404 for a missing asset in the console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@casper/nightshade-core",
-  "version": "2.18.3",
+  "version": "2.18.4",
   "description": "Casper pattern and style scss library. Under development.",
   "scripts": {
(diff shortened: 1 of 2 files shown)


victorwhy:
what is replacing this input arrow?

spanuska:
The arrow is already coming from someplace else (sprite file). Currently, there is an error in the console because it can't find this asset, but we don't need to find it because we're already getting it from somewhere else.

victorwhy:
yeah, sorry, wrote that before you responded to my comment in the channel

this looks good. make sure to add a release tag to the master after you merge this in, and update the release. let me know if you need some help with that

PR #195, opened 2017-03-27
Don't add special rules for webkit autofill

robwierzbowski:
Browser style has changed, making this override unnecessary.

With override: 

image

Without override:

image


ryanhagerty:
This looks familiar....I approve! DTM 🍀

robwierzbowski:
Also chat DTM from Joseph.

PR #196, opened 2017-08-04
Remove "1" prefix to tel: links, breaks int'l phone numbers

carpodaster:
Instead, `phone_number` should contain a valid, locally dialable phone number.

We already format the phone numbers properly in the static storefront. This `1` macro actually invalidates all of these `tel:` links currently, not just int'l phone numbers.

### Issues

Macros added `1` in front of every phone no. in `tel:` links. 

### Todos

Release NPM package after merge ^^

### Impacted Areas

### Steps to Reproduce

Ex.: casper.com/welcome, top right corner:

screen shot 2017-08-04 at 13 39 42

### DIFF
--- a/src/footer/footer.macros.html
+++ b/src/footer/footer.macros.html
@@ -32,7 +32,7 @@
 
       {% if not simple %}
       <div class="footer-nav-group footer-nav-group--right">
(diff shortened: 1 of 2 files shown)


edwardsfriedman:
For some reason I can't add the label but I'm dtm

carpodaster:
DTM'ed by @edwardsfriedman via email

PR #197, opened 2017-11-14
use git ssh url for nightshade icons

shuynh:
### Issues

What issues are being solved?

### Todos

What's outstanding?

### Impacted Areas

### Steps to Reproduce

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@casper/nightshade-core",
-  "version": "2.19.1",
+  "version": "2.19.2",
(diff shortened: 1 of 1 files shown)


victorwhy:
going to DTM this, but we should really fix the CI

victorwhy:
update this to the new version:

https://github.com/CasperSleep/nightshade-core/blob/351dc68f1a6bacc8780699cc157151e9547ba2f1/package.json#L3

shuynh:
casperbot u wild for dis 1

